PR #9, opened 2020-06-25
Create initial Circle config

LucasCairns:


.circleci/config.yml
```diff
@@ -0,0 +1,106 @@
...
+          name: Persist build number for deployment
+          command: |
+            mkdir -p /tmp/workspace/docker-build
+            cd /tmp/workspace/docker-build
+            echo preview > version-to-deploy.txt
+      - persist_to_workspace:
+          root: /tmp/workspace
+          paths:
+            - docker-build
+  build_drupal:
```
psoleckimoj, line 50: My only comment is maybe call it `build_prod` to make the job names consistent as they both build Drupal 😄

LucasCairns: Point taken, will update 😄

PR #24, opened 2020-07-02
Import Flysystem patch to migrate fields to Flysystem

psoleckimoj:
- From here https://www.drupal.org/project/flysystem/issues/2691731
- https://groups.drupal.org/node/518975
- https://github.com/cweagans/composer-patches


Dockerfile
```diff
@@ -25,6 +25,7 @@ RUN composer install \
 # Copy Project
 COPY modules/custom modules/custom
 COPY sites/ sites/
+COPY ./flysystem/FieldMigration.php /var/www/html/modules/contrib/flysystem/src/Form/FieldMigration.php
```
spikeheap: <I don't have an emoji for "I'm being really nitpicky"> I think we can remove the preceding `./` for consistency with the lines above

flysystem/FieldMigration.php
```diff
@@ -0,0 +1,414 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\flysystem\Form\FieldMigration.
+ */
```
spikeheap, line 6: Would it be worth us including the following?
- a link to the source of this patch
- license/credit to the author
- the purpose of the patch, and what it enables
- when it's safe to remove the patch (if ever)

psoleckimoj: @spikeheap still playing, not sure this is the right way to do it at the moment, you can apply patches using composer so messing with that atm :)

Dockerfile
psoleckimoj: Removed this line anyways :)

flysystem/FieldMigration.php
psoleckimoj: Now patching from Drupal website

PR #47, opened 2020-07-22
Add custom hostnames

spikeheap:
This PR:

- Adds a custom hostname for the frontend
- Refactors the ingress slightly to allow multiple hostnames


helm_deploy/prisoner-content-hub-backend/templates/ingress.yaml
```diff
@@ -1,23 +1,42 @@
 {{- if .Values.ingress.enabled -}}
-{{- $fullName := include "prisoner-content-hub-backend.fullname" . -}}
+{{- $fullName := include "prisoner-content-hub-frontend.fullname" . -}}
```
psoleckimoj, line 2: Are these supposed to be front-end?

spikeheap: 🤦

PR #50, opened 2020-07-22
Add hostname to ingress values

LucasCairns:
Added hostname to ingress values, this is used for generating the external file URL


helm_deploy/prisoner-content-hub-backend/values.production.yaml
```diff
@@ -1,4 +1,5 @@
 ingress:
+  hostName: cms-prisoner-content-hub-production.apps.live-1.cloud-platform.service.justice.gov.uk # added to support migration work, will need to find better solution
```
spikeheap, line 2: Should this be `manage.content-hub.prisoner.service.justice.gov.uk`?

PR #58, opened 2020-07-30
fixes to enable drupal cron jobs

davidthomas22:
Pairing @davidthomas22 and @spikeheap, we fixed up the cron jobs, testing the `prisoner-content-hub-development` namespace.

This PR:
- Fixes a naming issue with the cron job that caused the deployment to fail
- Improves the name of the pods for clarity
- Adds a security context to run the `curl` container non-root
- Corrects the `curl` image version to the latest one
- Runs the cron job every hour. It was running every 3 hours, but there's no reason not to run it more frequently

### DIFF
--- a/helm_deploy/prisoner-content-hub-backend/values.yaml
+++ b/helm_deploy/prisoner-content-hub-backend/values.yaml
@@ -75,7 +75,7 @@ ingress:
     nginx.ingress.kubernetes.io/proxy-body-size: 500m
 
 cron:
-  schedule: "*/* 3 * * *"
+  schedule: "0 * * * *"
   image:
-    repository: docker pull curlimages/curl
-    tag: 0.1
+    repository: curlimages/curl
+    tag: 7.71.1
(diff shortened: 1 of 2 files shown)


helm_deploy/prisoner-content-hub-backend/values.yaml
```diff
@@ -75,7 +75,7 @@ ingress:
     nginx.ingress.kubernetes.io/proxy-body-size: 500m
 
 cron:
-  schedule: "*/* 3 * * *"
+  schedule: "0 * * * *"
```
LucasCairns, line 78: Do we want to match what is currently scheduled in Drupal? `0 0/3 * * *` should match IIRC

LucasCairns: Just saw the PR comments... never-mind, makes sense

davidthomas22: I couldn't recall the rationale around the current schedule?

I can change to 0 */3 * * *

spikeheap: Unless there's a reason not to I think we should make it fairly frequent. The job should complete in seconds (it's a tiny database) so it just means the content will be indexed for search sooner, right?

LucasCairns: Makes sense to me, keep it frequent, less and often 👍

PR #59, opened 2020-07-30
Add prometheus alert config

davidthomas22:
-Configure Drupal search indexing alert

**NOTE**

Severity needs be defined once this has been requested and then provided by cloud platform team

 labels:
        severity: TODO


spikeheap:
This looks good 👍🙌. Once we've got the severity code we should test the full integration for this before we merge.

We can do that in the `development` namespace by:

1. Deleting the Drupal deployment (or scaling back to 0) so we have no pods to respond to the `curl` request.
2. Adjust the cron job to fire every minute, so we can see the alert pretty immediately.

PR #69, opened 2020-08-11
Update flysystem_s3, add pre-signed downloads

psoleckimoj:
- Removed flysystem_s3 v1
- Added flysystem_s3 v2
- Used updated patch for pre-signed downloads

### DIFF
--- a/helm_deploy/prisoner-content-hub-backend/values.yaml
+++ b/helm_deploy/prisoner-content-hub-backend/values.yaml
@@ -6,7 +6,7 @@ maxReplicaCount: 6
 
 image:
   repository: mojdigitalstudio/prisoner-content-hub-backend
-  pullPolicy: Always
+  pullPolicy: IfNotPresent
   tag: latest
 
 tier: backend
--- a/helm_deploy/prisoner-content-hub-backend/values.yaml
+++ b/helm_deploy/prisoner-content-hub-backend/values.yaml
@@ -6,7 +6,7 @@ maxReplicaCount: 6
 
 image:
   repository: mojdigitalstudio/prisoner-content-hub-backend
-  pullPolicy: IfNotPresent
+  pullPolicy: Always
   tag: latest
 
 tier: backend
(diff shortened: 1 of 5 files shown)


helm_deploy/prisoner-content-hub-backend/values.yaml
```diff
@@ -6,7 +6,7 @@ maxReplicaCount: 6
 
 image:
   repository: mojdigitalstudio/prisoner-content-hub-backend
-  pullPolicy: IfNotPresent
+  pullPolicy: Always
```
spikeheap, line 9: This is really handy for debugging & local dev, but we'll probably want to remove it in favour of specifying the version explicitly for CI (which makes rollbacks nice and fast because the image is already present). 

I wonder if there's a way to override the `pullPolicy` when deploying locally to the `development` namespace (which I'm guessing is the main/only time we'll be using the `latest` tag?

LucasCairns: We can add that to the environment yamls in Helm? we can revert the change in this PR and raise another for that specific behaviour?

LucasCairns: PR raised for this -> https://github.com/ministryofjustice/prisoner-content-hub-frontend/pull/48

LucasCairns: Raised a couple of draft PR's to support this behaviour
- https://github.com/ministryofjustice/prisoner-content-hub-frontend/pull/48
- https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/70

LucasCairns: Reverted the default pull policy

patches/flysystem_s3_2.0.0-rc1_enable_pre-signed_downloads.patch
```diff
@@ -176,7 +177,7 @@ index d9b5f26..6ebd2ef 100644
 +            'max-age' => 0,
 +          ],
 +        ];
-+        $this->renderer->renderRoot($build);
++        // $this->renderer->renderRoot($build);
```
spikeheap, line 180: https://www.drupal.org/project/flysystem_s3/issues/2772847#comment-13783915

PR #72, opened 2020-08-19
Scheduled publishing

spikeheap:
This PR adds the ability to schedule when a post will be published or unpublished, using the `drupal/scheduler` plugin. I spent an hour or so evaluating the plugin and testing it locally

Screenshot 2020-08-11 at 16 37 05

We also update the cron trigger to run more frequently (every 5 minutes). Scheduled posts are evaluated each `cron` run, so a post scheduled for `10:00:01` would _actually_ be published at `10:05:00` when the next job runs.

## Additional steps

Once this PR is merged we need to:

- [ ] enable the module in the [production Drupal UI](https://manage.content-hub.prisoner.service.justice.gov.uk/admin/modules).
- [ ] enable scheduled publishing for each content type in `Structure -> Content Types -> * -> Edit -> Scheduler`, enabling publish/unpublish and setting it to use an inline fieldset:
Screenshot 2020-08-19 at 08 49 12

- [ ] let people know about it (maybe with a Confluence page?)

I had wanted to set configuration in `settings.php`, but the nested, dynamic nature of applying this to each content type made it prohibitively difficult. Given more time I'd like to get `drush` working so we could apply these changes programmatically.

## Intent

Scheduled publishing, alongside the change in publishing workflow to make "Published" content immediately available, reduces the need for the content team to work over the weekend.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,7 @@
         "drupal/raven": "^2.27",
         "drupal/rest_menu_items": "^2.6",
         "drupal/restui": "^1.18",
+        "drupal/scheduler": "^1.3",
         "drupal/search_api": "^1.17",
         "drupal/token": "^1.5",
         "drupal/video": "^1.4",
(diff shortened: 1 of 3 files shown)


psoleckimoj (approved):
LGTM! It's like the onsite content teams dreams are all coming true at once 😀

PR #74, opened 2020-09-15
Allow filtering by prison

psoleckimoj:
- Create new function for checking prison id
- Accept prison id in api

### DIFF
--- a/modules/custom/moj_resources/src/ContentApiClass.php
+++ b/modules/custom/moj_resources/src/ContentApiClass.php
@@ -107,7 +107,7 @@ class ContentApiClass
   {
     $content_type = $node->type->target_id;
 
-    if ($prison_id !== 0 && !in_array($prison, $node->field_moj_prisons)) {
+    if ($prison !== 0 && !in_array($prison, $node->field_moj_prisons)) {
       return [];
     }
 
--- a/modules/custom/moj_resources/src/ContentApiClass.php
+++ b/modules/custom/moj_resources/src/ContentApiClass.php
@@ -115,16 +115,9 @@ class ContentApiClass
     $content_type = $node->type->target_id;
 
     if (($this->prison != 0) && (count($node->field_moj_prisons) > 0)) {
-      $found = false;
+      $prison_ids = array_column($node->field_moj_prisons, 'target_id');
 
-      foreach ($node->field_moj_prisons as $key => $n) {
-        if ($this->prison == $n->target_id) {
-          $found = true;
-          break;
-        }
-      }
-
-      if (!$found) {
+      if (!in_array($this->prison, $prison_ids)) {
         return [];
       }
     }
--- a/modules/custom/moj_resources/src/ContentApiClass.php
+++ b/modules/custom/moj_resources/src/ContentApiClass.php
@@ -115,23 +115,23 @@ class ContentApiClass
     $content_type = $node->type->target_id;
 
     if (($this->prison != 0) && (count($node->field_moj_prisons) > 0)) {
-      // $found = false;
+      $found = false;
 
-      // foreach ($node->field_moj_prisons as $key => $n) {
-      //   if ($this->prison == $n->target_id) {
-      //     $found = true;
-      //     break;
-      //   }
-      // }
+      foreach ($node->field_moj_prisons as $key => $n) {
+        if ($this->prison == $n->target_id) {
+          $found = true;
+          break;
+        }
+      }
 
-      // if (!$found) {
-      //   return [];
-      // }
-      $prison_ids = array_column($node->field_moj_prisons, 'target_id');
-
-      if(!in_array($this->prison, $prison_ids)) {
+      if (!$found) {
         return [];
       }
+      // $prison_ids = array_column($node->field_moj_prisons, 'target_id');
+
+      // if(!in_array($this->prison, $prison_ids)) {
+      //   return [];
+      // }
     }
 
     $defaults = $this->createItemResponse($node);
--- a/modules/custom/moj_resources/src/ContentApiClass.php
+++ b/modules/custom/moj_resources/src/ContentApiClass.php
@@ -115,9 +115,16 @@ class ContentApiClass
     $content_type = $node->type->target_id;
 
     if (($this->prison != 0) && (count($node->field_moj_prisons) > 0)) {
-      $prison_ids = array_column($node->field_moj_prisons, 'target_id');
+      $found = false;
 
-      if (!in_array($this->prison, $prison_ids)) {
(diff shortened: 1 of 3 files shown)


modules/custom/moj_resources/src/Plugin/rest/resource/ContentResource.php
```diff
@@ -156,6 +167,23 @@ protected function checkContentIdParameterIsNumeric()
...
+
+    protected function checkPrisonIsNumeric()
+    {
+        if (is_numeric($this->paramater_prison)) {
+            return true;
+        }
+        throw new NotFoundHttpException(
+            t('The prison parameter must be a numeric'),
+            null,
+            404
```
spikeheap, line 179: Should this be `400`/`BadRequestHttpException` to be consistent with the message on L177?

psoleckimoj: I just pulled the function from another `*Resource.php` file 😆

spikeheap:
This looks good, and I'm loving the prospect of having a single frontend deployment.

The $1million question is: can we introduce tests to validate these changes?

psoleckimoj:
> This looks good, and I'm loving the prospect of having a single frontend deployment.
> 
> The $1million question is: can we introduce tests to validate these changes?

It's a difficult one, e2e ideally to be able to check content that should be available across all sites vs that which is restricted etc I guess.

modules/custom/moj_resources/src/Plugin/rest/resource/ContentResource.php
psoleckimoj: If you look on the front-end you'll see it 404s when you can't view a piece of content so don't really want to change this until we know why it's a 404....lovely code we have 😆

psoleckimoj: Actually it'd be a very, very rare situation where this wouldn't be numeric since the `setPrison()` function pretty much ensures it is.

psoleckimoj:
Need to check this with current FE. The current HubClient passes prisonId with every request anyway so not sure what effect that will have now this code is using that arg.

spikeheap:
> Need to check this with current FE. The current HubClient passes prisonId with every request anyway so not sure what effect that will have now this code is using that arg.

It feels like this is a great opportunity to bite the bullet and introduce at least an e2e so we can feel more confident about the change. This change is pretty valuable, so I wonder if we could get buy-in for something in the next sprint to achieve this?

modules/custom/moj_resources/src/Plugin/rest/resource/ContentResource.php
spikeheap: This makes me a little uneasy, because the response we're introducing isn't "correct" even though it is consistent. If we caught `4xx`in the frontend and continued to show `404` to the user, would that allow us to retain the user-facing behaviour _and_ enable us to return more meaningful status codes from the Drupal API? It would also help us progressively improve the Drupal API to get us out of the technical debt we're in and make it easier/nicer to work with 🥳 .

I can see it being useful to capture these errors (with something like Sentry), because we shouldn't ever see a non-numeric prison, but a `404` for a piece of content is entirely predictable and wouldn't be an error state.

psoleckimoj: Aye, changing things like this needs proper time and understanding the code to make sure changing the response code doesn't break the frontend.

Totally agree on it being done properly....including renaming all the paramaters to parameters 😆

psoleckimoj:
@spikeheap changed to 400s and then to check specific content I used:

https://cookhamwood-prisoner-content-hub-development.apps.live-1.cloud-platform.service.justice.gov.uk/content/5238

This should be available in Berwyn and Wayland but not in Cookham Wood, it works as it should.

To test the no sites selected (so should work on al sites) I used:

https://wayland-prisoner-content-hub-development.apps.live-1.cloud-platform.service.justice.gov.uk/content/6291

Works on all sites.

These are the cases we need to check for 🙂

modules/custom/moj_resources/src/ContentApiClass.php
```diff
@@ -106,6 +113,22 @@ private function loadNodesDetails($nid)
...
+          $found = true;
+          break;
+        }
+      }
+
+      if (!$found) {
+        return [];
+      }
+    }
+
```
spikeheap, line 131: I wonder if we could use `in_array(thing, array)` here? 

Assuming `$node->field_moj_prisons` is an array, 

We can test this with:

> `var_dump` just prints the data type out to the console in `php -a`, which is handy for debugging.

```php
$prison_list = array(
  0 => array(
    'target_id' => '123',
    'name' => 'test 1'
  ), 
  1 => array(
    'target_id' => '234',
    'name' => 'test 2'
  )
);

# extracts a nested column into a 1-dimensional array
$prison_ids = array_column($prison_list, 'target_id');

# check for presence with in_array
var_dump(in_array('123', $prison_ids));
#> bool(true)

var_dump(in_array('432', $prison_ids));
#> bool(false)

var_dump(in_array(0, $prison_ids));
#> bool(false)

var_dump(in_array(null, $prison_ids));
#> bool(false)
```

I _think_ that means we can simplify the above to:
```php
$prison_ids = array_column($node->field_moj_prisons, 'target_id')

if(!in_array($this->prison, $prison_ids)) {
  return [];
}
```

spikeheap:
👍 this is looking good. I'd _love_ to figure out how to run some tests over this, but if it's the last thing blocking us switching to hostname identification of prison then we could card up the tests and look at prioritising them soon.

modules/custom/moj_resources/src/ContentApiClass.php
psoleckimoj: Looks good, made the change, will push to dev and check 👍

psoleckimoj: @spikeheap as with all things PHP this doesn't work as expected, the array functions are a bit weird.

psoleckimoj: Actually that could be a load of beep...I'll try it again later.

psoleckimoj: So yes, Drupal/PHP are playing weirdness here. I seem to remember I came across this on something else ages ago, the 'arrays' aren't arrays even though they look like arrays and everything says they should behave like arrays, there's something in the data structure that means PHP doesn't see them as such. So I've reverted it back to my long-winded way of doing it 😄

LucasCairns (approved):
LGTM :shipit:

davidthomas22 (commented):
LGTM

PR #75, opened 2020-09-16
Revert "Use namespace-specific nginx for non-production environments"

spikeheap:
Reverts ministryofjustice/prisoner-content-hub-backend#73

Cloud Platform have abandoned the custom-ingress approach, so we can undo this change.


psoleckimoj (approved):
LGTM

PR #78, opened 2020-09-25
Move API health endpoint to /health

psoleckimoj:
This may need some tweaks internally in the Drupal admin area.

### DIFF
--- a/helm_deploy/prisoner-content-hub-backend/values.yaml
+++ b/helm_deploy/prisoner-content-hub-backend/values.yaml
@@ -14,11 +14,11 @@ tier: backend
 application:
   port: 8080
   liveness:
-    endpoint: /api/health
+    endpoint: /health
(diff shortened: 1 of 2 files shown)


helm_deploy/prisoner-content-hub-backend/values.yaml
```diff
@@ -14,11 +14,11 @@ tier: backend
 application:
   port: 8080
   liveness:
-    endpoint: /api/health
+    endpoint: /
```
psoleckimoj, line 17: Note, these will be moved to `/health` once the change is in production. Needs to be set to this otherwise the pods never successfully pass a health check as the URL they check doesn't exist until the module has been removed and added back in the Drupal admin area.

davidthomas22 (approved):
LGTM

LucasCairns (approved):
LGTM

PR #80, opened 2020-10-02
Tidy up ordering of category featured content

psoleckimoj:
- Stop limiting promoted nodes early on since results are limited at the end
- Tidy up promoted terms functions

### DIFF
--- a/modules/custom/moj_resources/src/CategoryFeaturedContentApiClass.php
+++ b/modules/custom/moj_resources/src/CategoryFeaturedContentApiClass.php
@@ -132,25 +132,6 @@ class CategoryFeaturedContentApiClass
     return $result;
   }
 
-  private function promotedNodes($category, $prison)
-  {
-    $results = $this->entity_query->get('node')
-      ->condition('status', 1)
-      ->condition('field_moj_category_featured_item', 1)
-      ->accessCheck(false);
-
-    $results = getPrisonResults($prison, $results);
-
-    if ($category !== 0) {
-      $results->condition('field_moj_top_level_categories', $category);
-    };
-
-    $nodes = $results->execute();
-    $promotedContent = $this->loadNodesDetails($nodes);
-
-    return array_map(array($this, 'decorateContent'), $promotedContent);
-  }
-
   private function extractSeriesIdsFrom($nodes)
   {
     $seriesIds = [];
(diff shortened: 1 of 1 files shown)


modules/custom/moj_resources/src/CategoryFeaturedContentApiClass.php
```diff
@@ -139,19 +139,27 @@ private function extractSeriesIdsFrom($nodes)
...
     }
 
-    return $seriesIds;
+    return array_unique($seriesIds);
   }
 
   private function promotedSeries($category, $prison)
   {
-    $nids = $this->allNodes($category);
+    $results = $this->entity_query->get('node')
```
psoleckimoj, line 147: Used to be `allNodes` but makes easier to read as only used here.

modules/custom/moj_resources/src/CategoryFeaturedContentApiClass.php
```diff
@@ -139,19 +139,27 @@ private function extractSeriesIdsFrom($nodes)
       $seriesIds[] = $n->field_moj_series->target_id;
     }
 
-    return $seriesIds;
+    return array_unique($seriesIds);
```
psoleckimoj, line 142: Moved to the function rather than being run immediately on the results of calling this function.

modules/custom/moj_resources/src/CategoryFeaturedContentApiClass.php
```diff
@@ -164,42 +172,22 @@ private function promotedNodes($category, $number, $prison)
       $results->condition('field_moj_top_level_categories', $category);
     };
 
-    $results->range(0, $number);
```
psoleckimoj, line 167: Removed because with changed ordering you could theoretically have more than 8 matching tags that push all of this data out so should really get everything for tags and nodes, merge then sort..

modules/custom/moj_resources/src/CategoryFeaturedContentApiClass.php
```diff
@@ -164,42 +172,22 @@ private function promotedNodes($category, $number, $prison)
...
+        ($prison == $item->field_promoted_to_prison->target_id || !$item->field_promoted_to_prison->target_id)) {
         return true;
-      } elseif ($item->field_moj_category_featured_item->value == true && !$item->field_promoted_to_prison->target_id) {
-        return true;
-      } else {
-        return false;
       }
-    });
 
-    usort($promotedTerms, function ($a, $b) {
```
psoleckimoj, line 201: Unnecessary search as done later on.

modules/custom/moj_resources/src/CategoryFeaturedContentApiClass.php
```diff
@@ -164,42 +172,22 @@ private function promotedNodes($category, $number, $prison)
...
-
-    return $results->execute();
-  }
-
   private function promotedTerms($termIds, $prison)
   {
     $terms = $this->term_storage->loadMultiple($termIds);
     $promotedTerms = array_filter($terms, function ($item) use ($prison) {
-      if ($item->field_moj_category_featured_item->value == true && $prison == $item->field_promoted_to_prison->target_id) {
+      if ($item->field_moj_category_featured_item->value == true &&
```
psoleckimoj, line 185: Moved to single statement.

LucasCairns (approved):
👍

PR #87, opened 2020-10-30
Add filtering by Prison Type for Series content

LucasCairns:


modules/custom/moj_resources/src/SeriesContentApiClass.php
```diff
@@ -113,25 +115,22 @@ private function decorateSeries($node)
...
       $result["description"] = $curr->field_moj_description[0];
       $result["categories"] = $curr->field_moj_top_level_categories;
+      $result["prison_types"] = $curr->field_prison_types;
+
       if ($curr->field_moj_secondary_tags) {
         $result["secondary_tags"] = $curr->field_moj_secondary_tags;
       } else {
         $result["secondary_tags"] = $curr->field_moj_tags;
       }
-      $result["prisons"] = $curr->field_moj_prisons;
```
psoleckimoj, line 121: We can probably remove this from all decorators anyways, I think content used it when it did the access checks on the frontend but that's now in the backend :)

psoleckimoj (dismissed):
LGTM

modules/custom/moj_resources/src/SeriesContentApiClass.php
```diff
@@ -58,6 +59,7 @@ public function __construct(
     QueryFactory $entityQuery
   ) {
     $this->node_storage = $entityTypeManager->getStorage('node');
+    $this->term_storage = $entityTypeManager->getStorage('taxonomy_term');
```
psoleckimoj, line 62: Needs adding as a protected var further up, even if PHP doesn't seem to care 😆

LucasCairns: Added the protected member variable for TermStorage

psoleckimoj (dismissed):
LGTM!

modules/custom/moj_resources/src/Utilities.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+class Utilities
+{
+  public static function filterByPrisonTypes($prison_types, $query) {
+    $prison_type_or_group = $query
+      ->orConditionGroup()
+      ->condition('field_prison_types', $prison_types, 'IN')
+      ->condition('field_prison_types', NULL, 'IS NULL')
+      ->notExists('field_prison_types');
```
spikeheap, line 16: Does this match the logic in https://docs.google.com/document/d/1z2A_NxAy-dJ6_SmByUnSXIq2WKvmQxNqodZNbG0ZnNo/edit? It looks like it includes an item when the field is empty on L15/16), but in that case I'd expect that item to be excluded from the set. I might be reading this wrong though.

spikeheap:
This looks good. As it introduces business logic, I'd be much more comfortable if we added some sort of testing around this. Is it feasible to introduce a lightweight Cypress end-to-end test to increase our confidence in these changes?

modules/custom/moj_resources/src/Utilities.php
```diff
@@ -0,0 +1,34 @@
+<?php
```
psoleckimoj, line 1: Actually, thinking, do we not want to break this off into it's own PR? Then it can be merged and used by anything using it afterwards 👍

LucasCairns: We can, or depending on whether we merge to `main` or to a release branch it should be available

modules/custom/moj_resources/src/Utilities.php
LucasCairns: IIRC a prison type can be `null` for a piece of content, we already validate that the series has a matching prison type against the prison before executing the query

modules/custom/moj_resources/src/SeriesContentApiClass.php
LucasCairns: Yup, totally, we can remove this across the board as we update each endpoint

modules/custom/moj_resources/src/Utilities.php
spikeheap: :+1: for an episode it may be `null` if the series has the prison type set. Maybe the name of this function is what's off, as the logic for filtering by prison type is (normally) exclusive of `null`? 

This makes me wonder:
- Does this want to live in a utility function if it only applies to episodes in a series? 
- Thinking about making the business logic cohesive, should the function be aware of the series/episode relationship?

LucasCairns: But assuming no `prison_type` would be implicitly `global` would that mean this filter is valid for our use case?

spikeheap: From [Side note: making things less implicit](https://docs.google.com/document/d/1z2A_NxAy-dJ6_SmByUnSXIq2WKvmQxNqodZNbG0ZnNo/edit), if no `prison_type` is present, that means "no prison types" rather than "all prison types".

modules/custom/moj_resources/src/Utilities.php
```diff
@@ -0,0 +1,34 @@
...
+ * Utilities
+ */
+
+class Utilities
+{
+  public static function filterByPrisonCategories($prison_categories, $query) {
+    $prison_categories_or_group = $query
+      ->orConditionGroup()
+      ->condition('field_prison_categories', $prison_categories, 'IN')
+      ->condition('field_prison_categories', NULL, 'IS NULL') // this condition needs to remain in until we do a full launch otherwise no series will work
```
psoleckimoj, line 15: Added a comment but if we commit this as intended, i.e. this and the one below removed then Series as it stands won't work, either that or we go through Series and add this data as a test?

LucasCairns (commented):
Superseded by `feature/series-api-use-prison-category`, closing the PR

PR #89, opened 2020-11-05
Create Dockerfile to execute PHPUnit tests for custom modules

LucasCairns:


psoleckimoj (approved):
Like it! LGTM!

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM drupal:8.9.2-apache
+FROM drupal:8.9-apache AS test
```
spikeheap, line 1: Should we continue to track the patch version of the image, as we saw a change which broke for us between `8.9.2` and `8.9.3`?

LucasCairns: This PR fixes the issue we had with that particular breaking change, not impossible something like that could happen again, but breaking our build would certainly bring it to our attention

spikeheap: Sorry, hit "resolve" by mistake...

I think it's still sensible to pin to a patch version – that's the effect that `package.lock` has. The breakage we saw didn't show up in the Docker build (for me at least), so there's a risk that we could ship a build that broke production without us knowing about it 🙀.

Great work fixing that breakage so we can get up to 8.9.3 🙌. What was the problem?

Dockerfile
```diff
@@ -7,15 +7,54 @@ RUN apt-get update && apt-get install -y \
...
+
+FROM drupal:8.9-apache
+
+# Install Composer and it's dependencies
+RUN apt-get update && apt-get install -y \
+  curl \
+  git-core \
+  mediainfo \
+  unzip \
+  && rm -rf /var/lib/apt/lists/*
```
spikeheap, line 47: 🙌 loving the multi-stage Dockerfile ❤️ 

We can dry up the commonality between test and production builds using an interim step, e.g. `base`:

```Dockerfile
FROM drupal:8.9.2-apache as base
RUN apt-get update && apt-get install ...

FROM base as test
RUN rm -rf ..?* .[!.]* *
RUN composer create-project drupal-composer/drupal-project:8.x-dev . --stability dev --no-interaction

FROM base as production
RUN composer dump-autoload --optimize
```

Sorry if I'm double-commenting, but I thought I'd added this and can't see it...

Dockerfile
LucasCairns: It was where Drupal was installed in the image, the change wasn't documented, but set up a new `WORKDIR` to install our stuff in to the correct place

Dockerfile
LucasCairns: Updated the `Dockerfile` and pinned to `8.9.6`

Dockerfile
LucasCairns: pinned to 8.9.6

Dockerfile
```diff
@@ -46,6 +79,6 @@ RUN composer dump-autoload --optimize
 RUN composer clear-cache
 
 # Update permisions
-RUN chown -R www-data:www-data /var/www/html/
+RUN chown -R www-data:www-data /opt/drupal/web/
```
spikeheap, line 82: Now we're explicit about the `WORKDIR`, it feels like we can dry this up in one of two ways which might help keep things consistent in future:

1. Use `.`, here, so we're using the working directory
2. Set an environment variable, and use that to set this and `WORKDIR`

Dockerfile
```diff
@@ -7,16 +7,49 @@ RUN apt-get update && apt-get install -y \
...
+###########################################################################################
+# Create runtime image
+###########################################################################################
+
+FROM base
+
+WORKDIR /opt/drupal/web
+
+# Copy in Composer configuration
+COPY composer.json composer.lock /opt/drupal/web/
```
spikeheap, line 52: I've commented below about `WORKDIR` and keeping things DRY, and it applies here too.

Dockerfile
```diff
@@ -7,16 +7,49 @@ RUN apt-get update && apt-get install -y \
...
+
+RUN ../vendor/bin/phpunit -c core --testsuite unit --debug --verbose
 
+###########################################################################################
+# Create runtime image
+###########################################################################################
+
+FROM base
+
+WORKDIR /opt/drupal/web
```
spikeheap, line 49: The parent Dockerfile sets `WORKDIR` to be `/opt/drupal`. I'm interested in why it needs to be different for our project?

If we do need to change it, we can use `WORKDIR web` to shift _into_ the web dir. That might reduce the chance of a future change like the one we saw recently from breaking our build.

psoleckimoj (approved):
LGTM!

davidthomas22 (approved):
LGTM

Dockerfile
```diff
@@ -7,16 +7,49 @@ RUN apt-get update && apt-get install -y \
...
+RUN echo 'memory_limit = -1' > /usr/local/etc/php/php-cli.ini
+
+# Remove the vanilla Drupal ready to install a dev version
+RUN rm -rf ..?* .[!.]* *
+
+# Install Drupal 8.x Dev
+RUN composer create-project drupal-composer/drupal-project:8.x-dev . --stability dev --no-interaction
+
+# Update autoloads
+RUN composer dump-autoload --optimize
```
spikeheap, line 33: Do we need to `dump-autoload` for the tests to run?

Dockerfile
```diff
@@ -7,16 +7,49 @@ RUN apt-get update && apt-get install -y \
   mediainfo \
   unzip \
   && rm -rf /var/lib/apt/lists/*
-
 RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
 RUN php composer-setup.php --install-dir=/bin --filename=composer --version=1.10.16
 RUN php -r "unlink('composer-setup.php');"
```
spikeheap, line 12: 🙌 this is looking good. A few minor questions that might DRY things up.

Looking at https://github.com/docker-library/drupal/blob/b51cf402be28ed417827779e023a04da1888e8b0/8.9/apache-buster/Dockerfile#L66-L73 they have a nice one-step install/clear-out of Composer which might reduce our image step size. I know this isn't a change in this PR so it's not necessary for this one, but it's a good example of squashing install & cleanup steps into a single build step.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM drupal:8.9.2-apache
+FROM drupal:8.9.6-apache AS base
```
spikeheap, line 1: Not for this PR, but does this mean we can remove the Drupal version requirement from our composer setup? I'm interested in whether you're leaning more towards leaning on the Drupal Docker image or stepping back to the PHP/Apache base image in future iterations?

spikeheap (approved):
Looks good, thanks for the refactor 🙇‍♀️

psoleckimoj (approved):
LGTM! Liking the code, much tidier and easier to understand.

PR #90, opened 2020-11-12
Update SeriesAPI to take account of prison id and prison categories

psoleckimoj:


modules/custom/moj_resources/src/SeriesContentApiClass.php
```diff
@@ -195,44 +211,179 @@ private function translateNode(NodeInterface $node)
...
+      ->condition('field_prison_categories', $prison_categories, 'IN');
+    $no_prison_categories_filter = $query
+      ->andConditionGroup()
+      ->notExists('field_moj_prisons')
+      ->notExists('field_prison_categories');
+    $content_filter = $query
+      ->orConditionGroup()
+      ->condition('field_moj_prisons', $prison_id, '=')
+      ->condition($prison_categories_filter)
+      ->condition($no_prison_categories_filter);
```
spikeheap, line 319: Is this needed now we've back-filled categories to existing series?

modules/custom/moj_resources/src/SeriesContentApiClass.php
```diff
@@ -19,31 +20,42 @@ class SeriesContentApiClass
    *
    * @var array
    */
-  protected $nids = array();
+  protected $node_ids = array();
```
spikeheap, line 23: 🤩 loving the switch to more meaningful array names ❤️

LucasCairns: Definitely helps, there was a variable named `results` which turned out to be a `query`, something we should be mindful of as we go forward is fixing these variable names

psoleckimoj: It was a good call by @LucasCairns to do that 😄 We need to do it across all the code.

LucasCairns (commented):
Closing to change merge branch

PR #95, opened 2020-11-13
Feature/series api use prison category

LucasCairns:


modules/custom/moj_resources/src/SeriesContentApiClass.php
```diff
@@ -57,94 +69,96 @@ public function __construct(
...
-      $result["id"] = $curr->nid->value;
-      $result["image"] = $curr->field_moj_thumbnail_image[0];
-      $result["season"] = $curr->field_moj_season->value;
-      $result["episode"] = $curr->field_moj_episode->value;
-      $result["duration"] = $curr->field_moj_duration ? $curr->field_moj_duration->value : 0;
-      $result["description"] = $curr->field_moj_description[0];
-      $result["categories"] = $curr->field_moj_top_level_categories;
-      if ($curr->field_moj_secondary_tags) {
-        $result["secondary_tags"] = $curr->field_moj_secondary_tags;
+    return array_map(function ($node) {
```
psoleckimoj, line 119: Vars inside the fn need renaming to match `$seriesContent`

psoleckimoj (approved):
LGTM!

modules/custom/moj_resources/src/SeriesContentApiClass.php
```diff
@@ -168,80 +204,184 @@ function indexOf($comp, $array)
...
+    return $term;
+  }
+
+  /**
+   * Get Prison Categories for a Drupal term object
+   *
+   * @param EntityInterface $term
+   * @return int[]
+  */
+  private function getPrisonCategoriesForTerm($term) {
```
psoleckimoj, line 260: Maybe could change the name and remove references to Term? That way it can easily be a util function for content as well?

modules/custom/moj_resources/src/Utilities.php
```diff
@@ -0,0 +1,37 @@
...
+      ->orConditionGroup()
+      ->condition('field_moj_prisons', $prisonId, '=')
+      ->notExists('field_moj_prisons');
+
+    $filterByPrisonCategories = $query
+      ->orConditionGroup()
+      ->condition('field_prison_categories', $prisonCategories, 'IN')
+      ->notExists('field_prison_categories');
+
+    return $query
```
psoleckimoj, line 32: Need to ensure content owners tick the relevant prison categories that match the prison or none here if they want to filter just by prison. And the other way around if just selecting by prison category but logic looks good 👍

modules/custom/moj_resources/src/SeriesContentApiClass.php
LucasCairns: Updated 👍

modules/custom/moj_resources/src/Utilities.php
LucasCairns: Yup, agreed, being a utility function this is something we can iterate on across the board if needs be

PR #101, opened 2020-11-19
Refactor category featured content API endpoint

davidthomas22:
refactor variable names in API Class and content resource

### DIFF
--- a/modules/custom/moj_resources/src/CategoryFeaturedContentApiClass.php
+++ b/modules/custom/moj_resources/src/CategoryFeaturedContentApiClass.php
@@ -14,24 +14,12 @@ require_once('Utils.php');
 
 class CategoryFeaturedContentApiClass
 {
-  /**
-   * Node IDs
-   *
-   * @var array
-   */
-  protected $nodeIds = array();
-  /**
+   /**
    * Nodes
    *
    * @var array
    */
   protected $nodes = array();
-  /**
-   * Language Tag
-   *
-   * @var string
-   */
-  protected $lang;
   /**
    * Node_storage object
    *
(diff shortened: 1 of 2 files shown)


modules/custom/moj_resources/src/CategoryFeaturedContentApiClass.php
```diff
@@ -19,7 +19,7 @@ class CategoryFeaturedContentApiClass
    *
    * @var array
    */
-  protected $nids = array();
+  protected $nodeIds = array();
```
psoleckimoj, line 22: This is not actually used. A variable with the same name is used locally in one of the functions but not this private one.

modules/custom/moj_resources/src/CategoryFeaturedContentApiClass.php
```diff
@@ -83,53 +83,53 @@ protected function translateNode(NodeInterface $node)
...
   /**
-   * Get nids
+   * Get node ids
    *
    * @return void
    */
   protected function getFeaturedContentNodeIds($category, $number, $prison = 0)
   {
     $series = $this->promotedSeries($category, $prison);
     $nodes = $this->promotedNodes($category, $number, $prison);
```
psoleckimoj, line 93: In this context it'd make more sense to rename node to content as that's what it's pulling back.

modules/custom/moj_resources/src/CategoryFeaturedContentApiClass.php
```diff
@@ -83,53 +83,53 @@ protected function translateNode(NodeInterface $node)
...
-    $results = array_merge($series, $nodes);
+    $query = array_merge($series, $nodes);
```
psoleckimoj, line 94: ...and this actually is results and not a Drupal query. The queries are within the `promoted<blah>` functions. Can go through it with you tomorrow if you like?

modules/custom/moj_resources/src/CategoryFeaturedContentApiClass.php
```diff
@@ -208,28 +238,30 @@ private function promotedTerms($termIds, $prison)
...
    */
-  protected function serialize($item)
+  protected function serialize($term)
   {
-    $serializer = \Drupal::service($item->getType() . '.serializer.default'); // TODO: Inject dependency
-    return $serializer->serialize($item, 'json', ['plugin_id' => 'entity']);
+    $serializer = \Drupal::service($term->getType() . '.serializer.default'); // TODO: Inject dependency
+    return $serializer->serialize($term, 'json', ['plugin_id' => 'entity']);
   }
 }
```
LucasCairns, line 267: Do we use this method, if not we could remove?

psoleckimoj (approved):
LGTM

LucasCairns (approved):
LGTM 👍

PR #104, opened 2020-11-20
Refactor Term endpoint

psoleckimoj:


modules/custom/moj_resources/src/Plugin/rest/resource/TermResource.php
```diff
@@ -35,13 +35,6 @@
  *          type="integer",
  *          description="ID of term to return",
  *      ),
- *      @SWG\Parameter(
```
psoleckimoj, line 38: The language parameter is not actually used in the API.

LucasCairns: If it's not used you're right to remove it, it feels like if we could be in a position to generate the Swagger documentation again it would definitely help for it to be up to date 👍

psoleckimoj: Yeah, I noticed the translate function wasn't being run so there was no need for it.

modules/custom/moj_resources/src/TermApiClass.php
```diff
@@ -59,54 +32,39 @@ public function __construct(
     QueryFactory $entityQuery,
     Serializer $termSerializer
   ) {
-    $this->term_storage = $entityTypeManager->getStorage('taxonomy_term');
-    $this->entity_query = $entityQuery;
+    $this->termStorage = $entityTypeManager->getStorage('taxonomy_term');
     $this->termSerializer = $termSerializer;
```
LucasCairns, line 36: So we use the Drupal serialiser? we serialise our own objects with `createReturnObject`?

modules/custom/moj_resources/src/TermApiClass.php
```diff
@@ -14,33 +14,6 @@
...
-  protected $node_storage;
-  /**
-   * Entitity Query object
-   *
-   * @var Drupal\Core\Entity\Query\QueryFactory
-   *
-   * Instance of querfactory
-   */
-  protected $entity_query;
-
```
LucasCairns, line 43: Do we need a field definition for `$this->termStorage`?

psoleckimoj: It's used in `TermApiEndpoint`

modules/custom/moj_resources/src/TermApiClass.php
psoleckimoj: I actually don't know, it might have been coincidence in that something broke when I removed it 😆

modules/custom/moj_resources/src/TermApiClass.php
psoleckimoj: Resolved

LucasCairns (approved):
👍

PR #108, opened 2020-11-20
Refactor category menu endpoint

psoleckimoj:
- Refactored code
- Redid logic around returning results to simplify it

### DIFF
--- a/modules/custom/moj_resources/src/Plugin/rest/resource/CategoryMenuResource.php
+++ b/modules/custom/moj_resources/src/Plugin/rest/resource/CategoryMenuResource.php
@@ -108,6 +108,7 @@ class CategoryMenuResource extends ResourceBase
     $this->prisonId = self::setPrisonId();
     self::checkLanguageIdIsValid();
     self::checkCategoryIdIsNumeric();
+    self::checkPrisonIdIsNumeric();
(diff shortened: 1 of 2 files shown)


modules/custom/moj_resources/src/Plugin/rest/resource/CategoryMenuResource.php
```diff
@@ -79,48 +79,48 @@ class CategoryMenuResource extends ResourceBase
...
-    $this->parameter_prison = self::setPrison();
-    self::checkLanguageParameterIsValid();
-    self::checkCategoryIsNumeric();
-    parent::__construct($configuration, $plugin_id, $plugin_definition, $serializer_formats, $logger);
+    $this->availableLanguages = $this->languageManager->getLanguages();
+    $this->categoryId = self::setCategoryId();
+    $this->languageId = self::setLanguageId();
+    $this->prisonId = self::setPrisonId();
+    self::checkLanguageIdIsValid();
+    self::checkCategoryIdIsNumeric();
```
LucasCairns, line 110: Do we want to check if Prison ID is numeric?

psoleckimoj: Done 👍

LucasCairns (approved):
👍

PR #109, opened 2020-11-20
Refactor Related Content endpoint

psoleckimoj:


modules/custom/moj_resources/src/Plugin/rest/resource/RelatedContentResource.php
```diff
@@ -79,59 +79,59 @@ class RelatedContentResource extends ResourceBase
...
+        $this->prisonId = self::setPrisonId();
+        $this->numberOfResults = self::setNumberOfResults();
+        $this->offsetIntoNumberOfResults = self::setOffsetIntoNumberOfResults();
+        $this->sortOrder = self::setSortOrder();
+        self::checkLanguageIdIsValid();
+        self::checkCategoryIdIsNumeric();
         self::checkNumberOfResultsIsNumeric();
-        self::checkOffsetOfResultsIsNumeric();
-        parent::__construct($configuration, $plugin_id, $plugin_definition, $serializer_formats, $logger);
+        self::checkOffsetIntoNumberOfResultsIsNumeric();
```
LucasCairns, line 121: We should check Prison ID is numeric

LucasCairns (approved):
Just the one comment, but otherwise LGTM 👍

modules/custom/moj_resources/src/Plugin/rest/resource/RelatedContentResource.php
psoleckimoj: Done 👍

PR #111, opened 2020-11-20
Refactor new featured content resource endpoint

psoleckimoj:


modules/custom/moj_resources/src/NewFeaturedContentApiClass.php
```diff
@@ -57,104 +39,83 @@ public function __construct(
...
-   * @return void
-   */
-  private function getFeaturedContentNodeIds($number, $prison = 0)
+  public function FeaturedContentApiEndpoint($prisonId = 0)
   {
-    $results = $this->featuredNodes($number, $prison);
+    $results = $this->getFeaturedContent($prisonId);
 
-    return array_slice($results, 0, $number);
+    return array_slice($results, 0, 1);
```
psoleckimoj, line 55: I honestly cannot explain why this is required. Without it, it never returns.

LucasCairns (approved):
LGTM 👍

PR #113, opened 2020-11-21
Feature/add prison filtering to content

psoleckimoj:


modules/custom/moj_resources/src/Utilities.php
```diff
@@ -34,4 +38,88 @@ public static function filterByPrisonCategories($prisonId, $prisonCategories, $q
...
+  }
+
+  /**
+   * Get Prison Categories for a Drupal node object
+   *
+   * @param EntityInterface $term
+   * @return int[]
+  */
+  public static function getPrisonCategoriesFor($node) {
+    $prisonCategories = array();
```
psoleckimoj, line 109: Amusingly, `$a = [];` is not the same as `$a = array();`. The former causes `count()` to return `1` due to 

> When the parameter is neither an array nor an object with implemented Countable interface, 1 will be returned

modules/custom/moj_resources/src/ContentApiClass.php
```diff
@@ -24,79 +26,112 @@ class ContentApiClass
...
+   * Validate the content is OK for prison/prison categories
+   *
+   * @return NodeInterface
+   */
+  private function getMatchingContent() {
+    $prison = Utilities::getTermFor($this->prisonId, $this->termStorage);
+    $content = Utilities::getNodeFor($this->contentId, $this->nodeStorage);
+    $contentPrisons = Utilities::getPrisonsFor($content);
+
+    if (count($contentPrisons) === 0) {
```
LucasCairns, line 84: Could we use `empty($contentPrisons)`

modules/custom/moj_resources/src/Utilities.php
LucasCairns: Interesting, I'm assuming this is similar to in JS with `[1, 2, 3]` and `new Array(1, 2, 3)`?

modules/custom/moj_resources/src/ContentApiClass.php
psoleckimoj: Yep, it seems so.

modules/custom/moj_resources/src/Utilities.php
psoleckimoj: It caused me a lot of swearing when I could see the array was empty when printed out but still didn't pass the count.

modules/custom/moj_resources/src/ContentApiClass.php
psoleckimoj: @LucasCairns changed to empties...

modules/custom/moj_resources/src/Utilities.php
LucasCairns: I've also noticed similar issues with `Entity` objects returned by Drupal, they're iterable, but don't work with all Array functions

LucasCairns (approved):
👍

PR #114, opened 2020-11-24
Update Term resource to handle Prison ID

LucasCairns:


psoleckimoj (approved):
LGTM!

spikeheap:
Nice catch :). Would it be worth adding a test for this new (correct) behaviour?

PR #115, opened 2020-11-25
Refactor Vocabulary endpoint

psoleckimoj:
While we currently just use this for secondary tags, it's a generic endpoint that can return any taxonomy so we should add filtering.


LucasCairns (approved):
👍

PR #126, opened 2020-11-29
Add filtering to vocabulary

psoleckimoj:


modules/custom/moj_resources/src/Utilities.php
```diff
@@ -18,25 +18,35 @@ class Utilities {
     *
     * @param int $prisonId
     * @param int[] $prisonCategories
-    * @param QueryInterface
+    * @param QueryInterface $query
+    * @param bool $isTerm
```
psoleckimoj, line 22: If you try and query on a non-existent field the query blows up, added this switch.

modules/custom/moj_resources/src/Utilities.php
```diff
@@ -93,7 +103,7 @@ public static function getPrisonsFor($node) {
     $prisons = array();
 
     foreach ($node->field_moj_prisons as $prison) {
-      array_push($prisons, $prison->target_id);
+      array_push($prisons, intval($prison->target_id));
```
psoleckimoj, line 106: After finding some of these are coming through as strings when doing the Term API I don't trust these.

LucasCairns: So this does casting similar to `(int) $prisonCategory->target_id`?

LucasCairns (approved):
👍

modules/custom/moj_resources/src/Utilities.php
psoleckimoj: Yep, it's basically the same as `Number.parseInt()` in JS.

PR #136, opened 2020-12-02
Introduce Drush (and update a bunch of dependencies 🙈)

spikeheap:
### Description

> What changes are included in the PR. How do they affect user behaviour, technical debt, etc.
> Link to the Trello card this work relates to (or describe why a card isn't appropriate), e.g. "Fixes (Trello card URL)".

Trello cards: 
- https://trello.com/c/m48bNKes/1724-get-drush-working-so-we-can-programmatically-reinstall-modules-and-do-data-migrations.
- https://trello.com/c/eBTX67RF/63-update-drupal-and-dependencies

This PR introduces Drush, a PHP/Drupal task runner:

Screenshot 2020-12-02 at 13 54 39

### Intent

> Why do we want to make this change?

We've got a few use-cases for Drush:

- Reinstalling modules as part of a Helm deployment
- Upgrading Drupal
- Clearing caches and developing locally
- Data migrations, such as updating prison categories. This is the big one that'll help us loads as we try to simplify the information architecture.

### Considerations

> What do we need to think about when reviewing this PR? Are there special steps to consider when merging or deploying?
It's definitely worth checking a local build/run of this. I've tested it in Docker Compose, but not in other environments.

### Checklist

- [ ] I have made corresponding changes to the documentation.
  - if we like the look of this PR I'll add some bits to Confluence (or maybe the readme on this repo 🤷‍♂️, what do you think?)
- [ ] I have added tests that prove my fix is effective or that my feature works.
  - I've manually tested this, and the build itself will is tested by the CI Docker build.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
         "drupal/video": "^1.4",
         "drupal/videojs": "^1.0",
         "drupal/view_unpublished": "^1.0",
+        "drush/drush": "^10.3",
         "mhor/php-mediainfo": "^5.1.0"
     },
     "conflict": {
@@ -88,5 +89,8 @@
             "type": "composer",
             "url": "https://packages.drupal.org/8"
         }
-    ]
+    ],
+    "require-dev": {
+        "drupal/core-dev": "^8.9"
+    }
 }
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "74f21b5a210ddcc214ac165c346076ef",
+    "content-hash": "7341fe191ad4524f67bc8ca66f8ef165",
     "packages": [
         {
             "name": "asm89/stack-cors",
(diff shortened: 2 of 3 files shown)


spikeheap:
We often see `composer` use all its available memory and fail. To get around this we can use:

```
export COMPOSER_MEMORY_LIMIT=-1 
```

From https://github.com/drupal-composer/drupal-project/blob/9.x/composer.json, we need to add `drupal-dev` before we can get `drush` to work:

```
composer require --dev drupal/core-dev ^8.9
```

This fails with an error because no compatible versions can be identified. To work around this I used `--no-update` to break the installation down into a few steps:

```
composer require --dev drupal/core-dev ^8.9 --no-update
composer require drush/drush--no-update
composer update
```

### Running a database update

We need to update the db after a Drupal update (e.g. 8.9.6 )
```
./vendor/drush/drush/drush updatedb
./vendor/drush/drush/drush cache:rebuild
```


After all this we can reinstall modules using:

```
./vendor/drush/drush/drush pm:uninstall moj_resources
./vendor/drush/drush/drush pm:enable moj_resources
```

spikeheap:
From https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/136/commits/dede21ff5f42f087cb2df6d38efbfe077c4687b5: 

> Update dependencies and install drupal/dev and Drush... read on...  …
> I know what you're thinking, this should have been split into a couple of commits. This was stymied by my lack of
> composer knowledge and a couple of challenges:
> 
> - From https://github.com/drupal-composer/drupal-project/blob/9.x/composer.json, we need to add  before we can get  to work
> - Installing either of these fails because symfony is locked to 5.1.7. I found a recommendation to split the require and update steps, which worked
> here: https://www.drupal.org/project/drupalmoduleupgrader/issues/3072073.
> 
> So we ended up with the following commands:
> 
> composer require --dev drupal/core-dev ^8.9 --no-update
> composer require drush/drush--no-update
> composer update
> 
> Which is the entirety of this commit. I am sorry to my future self and any reviewers that this includes a bunch of dependency upgrades
> in the same place, but I lost 30 minutes trying to unpick it and didn't get anywhere. Brute Force for the win.

psoleckimoj (approved):
Gonna LGTM :)

PR #139, opened 2020-12-11
Amend sort functionality on series

davidthomas22:
*Add sort by drupal date field and updated date
*Filter content with episode number
*Filter content without episode number
*Sort series by date and then episode number


modules/custom/moj_resources/src/SeriesContentApiClass.php
```diff
@@ -101,9 +101,16 @@ public function SeriesNextEpisodeApiEndpoint($lang, $series_id, $number, $episod
   private function decorateSeries($node)
   {
     $results = array_reduce($node, function ($acc, $curr) {
-      $episode_id = ($curr->field_moj_season->value * 1000) + ($curr->field_moj_episode->value);
+      $episode_id = 0;
```
LucasCairns, line 104: We will want to use `camelCase` for variables to be consistent with code styling

modules/custom/moj_resources/src/SeriesContentApiClass.php
```diff
@@ -148,10 +167,22 @@ private function sortSeries(&$series, $sort_order)
         return ($a['episode_id'] < $b['episode_id']) ? -1 : 1;
       }
 
-      return ($b['episode_id'] > $a['episode_id']) ? 1 : -1;
+      return ($a['episode_id'] < $b['episode_id']) ? 1 : -1;
     });
 
-    return $series;
+    usort($no_episode, function ($a, $b) use ($sort_order) {
+      if ($a['episode_id'] == $b['episode_id']) {
```
LucasCairns, line 174: Do we need this? given this sort function uses content without an episode number

modules/custom/moj_resources/src/SeriesContentApiClass.php
```diff
@@ -205,9 +236,9 @@ private function getSeriesContentNodeIds($series_id, $number, $offset, $prison)
       ->condition('status', 1)
       ->accessCheck(false);
 
-    if ($series_id !== 0) {
+    // if ($series_id !== 0) {
```
LucasCairns, line 239: We should remove this instead of commenting out, ideally, if we intend to keep, would it be save to assume we should return a `400` response if there is no series ID?

modules/custom/moj_resources/src/SeriesContentApiClass.php
```diff
@@ -137,9 +144,21 @@ private function decorateSeries($node)
...
+      return $node["episode_id"] === 0;
+    });
+
+    $has_episode = array_filter($series, function($node) {
+      return $node["episode_id"] !== 0;
+    });
+
+
+
+    usort($has_episode, function ($a, $b) use ($sort_order) {
```
LucasCairns, line 161: I'm wondering if a `switch` would work better here? i.e
```
switch($sort_order) {
  case: 'ASC'
    return ...
  case: 'DSC'
    return ...
  default:
    return 0;
}
```
it would maybe make it easier to extend and make the sort order more explicit when reading the code
}
```

File: modules/custom/moj_resources/src/SeriesContentApiClass.php (hunk 2 above)
Comment (psoleckimoj) [reply modules/custom/moj_resources/src/SeriesContentApiClass.php:174]: It should be using `date` as the comparison and not `episode_id`

File: modules/custom/moj_resources/tests/src/Unit/SeriesContentApiClassTest.php [hunk 5]
```diff
@@ -53,6 +53,8 @@ public function testGetSeriesContentVideoFormat() {
       $this->assertEquals($content["season"], 1);
       $this->assertEquals($content["episode"], 1);
       $this->assertEquals($content["episode_id"], 1001);
+      $this->assertEquals($content["date"], "field_moj_date");
+      $this->assertEquals($content["last_updated"], "changed");
```
psoleckimoj, line 57: We're not using `last_updated` so it can go, same for the other change below.

LucasCairns (approved):
👍

PR #149, opened 2021-01-16
Add prison tagging action

swilson-mojhub:
- Add custom action module
- Add install and scheme info
- Add a custom action to tag content will all prisons

### Context

https://trello.com/c/64dZUJsD/1852-spike-drupal-vbo

### Intent

Screenshot 2021-01-16 at 11 19 41

Adds a custom action to the content admin to tag content with all prisons. 

_Note it only tags content if there are no prisons currently checked._

### Considerations

An update to the content admin to allow the content to filtered by prison tags would add value b/c it would reduce the number of content to be tagged, therefore reducing the load on the CMS

The module can be installed using `drush en custom_action_prison_tagging` Ther are no DB updates with this module.

### Checklist

- [X ] This PR contains **only** changes related to the above card
- [ ] Tested in Development

### DIFF
--- a/modules/custom/custom_action_prison_tagging/config/install/system.action.node_prison_tag_action.yml
+++ b/modules/custom/custom_action_prison_tagging/config/install/system.action.node_prison_tag_action.yml
@@ -1,0 +1,10 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - node
+id: node_prison_tag_action
+label: 'Tag content with prisons'
+type: node
+plugin: node_prison_tag_action
+configuration: {  }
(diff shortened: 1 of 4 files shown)


spikeheap (approved):
:+1: reviewed on a call with @swilson-mojhub. This PR is a good example of using a custom action, which we'll almost certainly want to use again in the near future.

A couple of things came up:

1. Ideally we'd pull prisons directly from Drupal rather than hard-coding them. This is a one-off action, so provided we remove this code once it's been run it's not worth the extra effort. :shipit: .
2. For future one-off data migrations we'll use update hooks on a module. This should give us a nice, clean history, and a library of previous migrations to build new ones from.

So, the plan is:
1. We'll merge this PR 🙌 ❤️ 
2. We'll run the migration through the UI
3. We'll revert this PR to remove the button & code

PR #152, opened 2021-01-21
Bugfix/vocabulary endpoint filtering

LucasCairns:
### Context

https://trello.com/c/l6Rr4q7X/1870-all-topics-page-categories-secondary-tags

- Visit `Browse All Topics` page for a prison
- Only relevant `Secondary Tags` should be displayed for that prison

Note: this can be compared against Production which is not affected by this issue

### Intent

- Taxonomies fetched via the `/api/v1/vocabulary/{type}` endpoint in Drupal should not be filtered by `Prison Category` unless it is a `Series`
- When validating a `Secondary Tag` has content (in order to be displayed for the `All Topics` page) we use `Prison Category` filtering

N/A

### Considerations

N/A

N/A

### Checklist

- [x] Tested in Development

### DIFF
--- a/modules/custom/moj_resources/src/VocabularyApiClass.php
+++ b/modules/custom/moj_resources/src/VocabularyApiClass.php
@@ -9,8 +9,6 @@ use Drupal\Core\Entity\EntityTypeManager;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Entity\Query\QueryFactory;
 
-require_once('Utils.php');
-
 /**
  * PromotedContentApiClass
  */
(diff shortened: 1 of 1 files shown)


spikeheap (approved):
I've not validated this change locally, but this seems reasonable :+1:.

@LucasCairns did you intend to include the update to `getAllSecondaryTagItemsFor($termId)`? Is that relevant to this change, and the method is just (historically) poorly named?

LucasCairns:
Totally meant to include that change, it checks secondary tags have content, ensuring that all tags on the All Topics page go somewhere. It's far from ideal code, but just fixing up at this point

PR #159, opened 2021-01-27
1892 enable xdebug

leonk:
### Context

https://trello.com/c/FbBnFGtj/1892-enable-xdebug

### Intent

Enable xdebug on local environments only.

I wanted this be done with zero-configuration (it should be enabled by default).  And only on local environments, so to achieve this I used the docker multi-stage environments to create a new "local" environment (where additional local only stuff can also be added to).  But to do this I've had to give the original environment a new name, and update the build task in the makefile.

### Considerations

This depends on PR https://github.com/ministryofjustice/prisoner-content-hub/pull/751

Note that the docker image built in circle has now changed (`FROM base as build`)

### Checklist

- [ ] Tested in Development


Dockerfile
```diff
@@ -75,3 +75,10 @@ RUN chmod u-w sites/default/settings.php \
 RUN chown -R www-data:www-data ./
 
 USER www-data
+
+FROM build as local
+USER root
+RUN pecl install xdebug-2.9.8 \
+  && docker-php-ext-enable xdebug
+USER www-data
+
```
spikeheap, line 84: This looks good! The default build artefact this way is a local dev build, which it’s safer to make explicit. 

If we add a single line to the end of this file like
‘’’
FROM build as production
‘’’

That should do it

LucasCairns (approved):
👍

PR #160, opened 2021-01-29
1893 move docroot

leonk:
### Context

https://trello.com/c/zmPOhBv2/1893-move-drupal-document-root

### Intent

Move Drupal into new docroot, to follow best practises.

### Checklist

- [ ] This PR contains **only** changes related to the above card
- [ ] Tests have been added/updated to cover the change
- [ ] Documentation has been updated where appropriate
- [ ] Tested in Development

PR #161, opened 2021-01-29
1898 fix composer file permissions

leonk:
### Context

https://trello.com/c/9syUElat/1898-unable-to-run-composer-inside-docker

### Intent

Fixes an issue with the composer cache directory.  Other changes are in https://github.com/ministryofjustice/prisoner-content-hub/pull/753

### Considerations

This PR is based off https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/160 (so the other should be merged in first).  They both modify the Dockerfile so didn't want to deal with any conflicts.

### Checklist

- [ ] Tested in Development


spikeheap (dismissed):
:+1: worth checking Apache ownership guidance, but this doesn't look like it'll affect any other files

leonk:
@spikeheap did a bit of a Google, found a lots of S/O questions, but these were mainly around handling multiple users and virtualhosts.
The apache docs just has this section about file ownership http://httpd.apache.org/docs/2.4/misc/security_tips.html#serverroot which basically says the main thing to be concerned about is the www-data user having access to files that the root user can run.  But it says that root never uses any files inside of htdocs (which is essentially /var/www) so I think we are good here.

spikeheap (approved):
:+1: Amazing, thanks for digging into that @leonk

PR #164, opened 2021-02-04
1900 get phpunit tests running on local docker environment

leonk:
### Context

https://trello.com/c/fIMUtXBN/1900-get-phpunit-tests-running-on-local-docker-environment

Note this PR is just about getting testing setup on local environments.  It's not intended (yet) to work on circleci.

### Intent

This PR makes several modifications to the `Dockerfile`, mostly are around the multistage builds, `local` is now built on top of `test` to allow testing to be done on local environments. I'm assuming this is how we want to work, and not have separate local docker containers for testing, as switching between the two could be time consuming.

Also the `phpunit.xml` file has been updated from the latest provided by Drupal core.

And the `makefile` now contains commands to run the tests.  You can do so by:
```
docker exec -it prisoner-content-hub_drupal_1 make run-tests
```

### Considerations

You will need the environment variables (copied into to your drupal.env file) from https://github.com/ministryofjustice/prisoner-content-hub/pull/755

### Checklist

- [x] Tested in Development


Dockerfile
```diff
@@ -18,52 +18,19 @@ RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" \
...
-RUN composer create-project drupal-composer/drupal-project:8.x-dev . --stability dev --no-interaction
-
-COPY phpunit.xml web/core/phpunit.xml
-COPY docroot/modules/custom web/docroot/modules/custom
-
-RUN vendor/bin/phpunit -c web/core --testsuite unit --debug --verbose
-
 ###########################################################################################
 # Create runtime image
 ###########################################################################################
```
spikeheap, line 24: Is this comment still valid? It looks like it's tied to `FROM base as build` which we're removing.

Dockerfile
```diff
@@ -72,18 +39,53 @@ COPY config/ config/
...
 RUN chown -R www-data:www-data /var/www
 
 USER www-data
 
-FROM build as local
+FROM test as local
 USER root
 RUN pecl install xdebug-2.9.8 \
   && docker-php-ext-enable xdebug
 USER www-data
```
spikeheap, line 68: This is nifty. Would we ever want to use `debug` within a test? If so we could add this to the `test` image above.

Dockerfile
```diff
@@ -72,18 +39,53 @@ COPY config/ config/
...
+
+# Change ownership of files
+RUN chown -R www-data:www-data ./
+RUN chown -R www-data:www-data /var/www
+
+USER www-data
+
 # Make build (and not local) the default target.
 # By ensuring this is the last defined target in the file.
 FROM build as production
```
spikeheap, line 91: As `build` is now the last step in the `Dockerfile`, we could remove these lines, and (maybe) rename L73 to `production` or `optimised` or something.

spikeheap (approved):
Looks good :+1:, just some minor nits

Dockerfile
leonk: No we wouldn't want to install xdebug on the test build (if that's what you mean), as it really slows things down.  You can only really use it on local environments.

Dockerfile
leonk: Good idea. I think `optimised` works well.  I've updated this now in the PR.

Dockerfile
leonk: I've updated the comment to say "Copy repository files".  Ideally I want to keep all the COPY commands in this section, so they're all in one place.

leonk:
Circleci gets this error on the build (when running `composer install`)
```
  [RuntimeException]                                               
  /opt/drupal/web/vendor does not exist and could not be created.  
```
https://app.circleci.com/pipelines/github/ministryofjustice/prisoner-content-hub-backend/573/workflows/10c2a95c-d764-4852-b411-dc25701e97a0/jobs/490/parallel-runs/0/steps/0-103

Assuming this is caused by composer being run from the `www-data` user, but can't see why this would be happening as that user is only switched to in the `test` and `local` builds.

leonk:
Fixed the composer issue error by bumping up the docker version number.  Previously we weren't specifying a version, and were getting 17.06.0 (not sure why exactly it was this specific version).  Under this version _all_ of the commands in the Dockerfile were being run (i.e. it wasn't skipping build steps that were related to other targets).

The current latest version we can switch to is 19.03.13 according to this page: https://circleci.com/docs/2.0/building-docker-images/
Using this version means that Docker runs the multistage build correctly.

leonk:
Updated Confluence docs: https://dsdmoj.atlassian.net/wiki/spaces/HUB/pages/2543157321/Testing+PHP+code

spikeheap (approved):

PR #165, opened 2021-02-09
1905 get phpunit tests running on circle ci

leonk:
### Context

https://trello.com/c/Z018ARHZ/1905-get-phpunit-tests-running-on-circle-ci

### Intent

Make necessary modifications to allow tests to run in circleci.  Apologies for the rather large PR, but all of this is necessary to get it to work.  
Also there are lots of commits in this PR, as I had to push changes to test in circleci.  So rather than going through each commit, it might be easier if l break down the major changes here:

#### Dockerfile
No longer using the dockerhub drupal base image, as this does a few things we don't want.  Mainly [composer create-project](https://github.com/docker-library/drupal/blob/master/8.9/php7.4/apache-buster/Dockerfile#L73).  This results in two versions created of the Drupal codebase (as we run `composer install` later on, which installs to a different directory).
To fix this, I copy-pasted the [first part of the Dockerfile](https://github.com/docker-library/drupal/blob/master/8.9/php7.4/apache-buster/Dockerfile#L1-L63) into our own Dockerfile.
I did look into a way of modifying the dockerhub image with multistage builds, that way we could still use this as our base, but not have to include the whole thing.  I forked the repo and made the necessary amendments : https://github.com/leonk/drupal-docker-multistage-builds. However, getting these onto dockerhub is not a straightforward process (which is documented [here](https://github.com/docker-library/faq#an-images-source-changed-in-git-now-what)).

#### Added a run_tests step to cirecleci.yml
Main thing to note here is I am running docker-compose inside cirecleci (as opposed to just using circleci's own docker support to create the two required images).  The main reason for this was to mimic how our normal environment is built, and also a bit of a lack of knowledge of circleci from my part.

#### Drupal changes
A few modifications in Drupal were required:
- Switching installation profile
- - The reasons for this is described in the [README file](https://github.com/ministryofjustice/prisoner-content-hub-backend/blob/f1bc8af0035d0b8fc44748d1b23c0373f2794188/docroot/profiles/prisoner_content_hub_profile/README.md)
- - There is an update hook to switch profiles, see https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/165/commits/fe963372294d00a8b4ca87fcc392e763a6da53c1 this needs quite a bit of testing before it gets to production (as switching profiles is technically not a supported thing you can do in Drupal).
- Removing custom module config
- - There was a bunch of configuration that came in our custom modules (inside `config/install` directories).  These would have no effect on the site unless you are installing the module again.  On the test environment we are installing these modules for the first time, so the config gets imported.  There were several errors with the config so it was easier to just remove them.

### Checklist

- [x] Tested in Development


.circleci/config.yml
```diff
@@ -143,6 +191,9 @@ workflows:
   version: 2
   build-test-deploy:
     jobs:
+      - run_tests:
+          <<: *feature_branch
```
spikeheap, line 195: Could this run for both the main branch as well as feature branches?

Dockerfile
```diff
@@ -17,12 +83,13 @@ RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" \
 
 # Set Timezone
 RUN echo "date.timezone = Europe/London" > /usr/local/etc/php/conf.d/timezone_set.ini
+# Set no memory limit (for PHP running as cli only).
+RUN echo "memory_limit = -1" > /usr/local/etc/php/conf.d/memory_limit.ini
```
spikeheap, line 87: The comment above mentions this only applies to the CLI. How is that ensured? Is the option only honoured for CLI use?

spikeheap (commented):
This PR contains a bunch of profile-switching/renaming changes as well as the CI bits. Is that intentional?

One nitpick is that the Docker Compose setup runs before we build the image, so we end up building twice. This shouldn't be the end of the world because we have layer caching, and we can always refactor the build steps later.

Dockerfile
leonk: Updated the file to be `/usr/local/etc/php/php-cli.ini` this only gets picked up by php running in cli.

.circleci/config.yml
leonk: Updated the config so that it should now run on all branches

leonk: And yes going back to your first comment.  The profile switching is required to install Drupal on circleci (it's explained in the new profiles README file).

spikeheap (approved):
:+1:

PR #167, opened 2021-02-22
1930 enable gds theme switcher

leonk:
### Context

https://trello.com/c/Xmv4dEIc/1930-enable-gds-theme-theme-switcher

### Intent

Enables the govuk_design_system theme, and adds ability for user to switch to this theme (via adding a user role to their account).  Each user will have to opt in to theme (i.e. they won't get it automatically).  They also have the option to then remove the role (and revert back to the default theme).

### Considerations

No modifications have been made to the GDS theme, so we're using it "as is".

### Checklist

- [ ] Tested in Development


config/sync/govuk_design_system.settings.yml
```diff
@@ -0,0 +1,15 @@
+phase_banner: ''
+phase_banner_message:
+  value: "<p>This is a new service – your <a class=\"govuk-link\" href=\"/contact/feedback\">feedback</a> will help us to improve it.</p>\r\n"
```
spikeheap, line 3: That feedback URL was dead for me. Should we link to the public Slack channel (or use a `mailto:`) as an interim?

spikeheap (approved):
👍 looks sensible, one minor comment

config/sync/govuk_design_system.settings.yml
leonk: We're not actually using the phase banner, the `phase_banner: ''` line means it's turned off. (We're not in an Alpha/Beta phase so doesn't make sense to have one).
That text value is just what comes with the theme, it's the default value.

spikeheap: :+1:, `''` is wonderfully intuitive for turning something off 🤪. Thanks, this is fine in that case 🙌

PR #169, opened 2021-03-03
1929 implement prison category content access rule

leonk:
### Context

https://trello.com/c/W51gad41/1929-implement-prison-category-content-access-rule

### Intent

This PR builds onto of https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/168 using the new prison context (passed via jsonapi url) to implement the prison category access rules.

It also adds in testing based on https://gitlab.com/weitzman/drupal-test-traits
We now test on the fully installed site, with all our config.  (Whereas previously a new "minimal" Drupal site would be spun up to test the modules in isolation).

### Considerations

It would be worth going over the tests in `PrisonerHubQueryAccessTest.php` to check they cover the requirements and different use cases (and if there's anything that has been missed).

### Checklist

- [x] Tested in Development


docroot/modules/custom/prisoner_hub_entity_access/src/EventSubscriber/QueryAccessSubscriber.php
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * The prison category field name.
+   *
+   * @var String
+   */
+  protected $prisonCategoryFieldName;
+
+  /**
+   * The prison category field name.
```
spikeheap, line 42: Typo?
```suggestion
   * The prison field name.
```

docroot/modules/custom/prisoner_hub_entity_access/src/EventSubscriber/QueryAccessSubscriber.php
```diff
@@ -0,0 +1,178 @@
...
+    $this->routeMatch = $route_match;
+    $this->prisonFieldName = $prison_field_name;
+    $this->prisonCategoryFieldName = $prison_category_field_name;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents() {
+    // TODO: Do we want to run this for Taxonomy Terms?  Or other entity types?
```
spikeheap, line 59: Is this `TODO` still outstanding?

docroot/modules/custom/prisoner_hub_entity_access/src/EventSubscriber/QueryAccessSubscriber.php
```diff
@@ -0,0 +1,178 @@
...
+   *   The current prison taxonomy term (loaded via the url).
+   *
+   * @return \Drupal\entity\QueryAccess\ConditionGroup
+   *   The condition group, to be added to the entity query.
+   */
+  protected function getPrisonConditionGroup(TermInterface $current_prison) {
+    $condition_group = new ConditionGroup('OR');
+    $condition_group->addCondition($this->prisonFieldName, $current_prison->id());
+    $bundles = $this->getFieldBundles('node', $this->prisonFieldName);
+    $condition_group->addCondition('type', $bundles, 'NOT IN');
```
spikeheap, line 112: Sorry, noob question: what are `FieldBundles` and what does this `OR` group do?

docroot/modules/custom/prisoner_hub_entity_access/tests/src/ExistingSite/PrisonerHubQueryAccessTest.php
```diff
@@ -0,0 +1,228 @@
...
+        ],
+        $this->prisonCategoryFieldName => [
+          ['target_id' => $this->anotherPrisonCategoryTerm->id()],
+        ],
+      ];
+      $this->createNode($values)->uuid();
+    }
+
+    $this->assertJsonResponseNodes($nodes_to_check);
+  }
```
spikeheap, line 205: Should we add a test for when content is tagged the same category but a different prison (asserting that it's not shown)?

docroot/modules/custom/prisoner_hub_prison_context/prisoner_hub_prison_context.info.yml
```diff
@@ -0,0 +1,8 @@
+name: 'Prisoner Hub Prison Context'
+type: module
+description: 'Provides a Drupal service for loading in the current prison context.'
```
spikeheap, line 3: Just thinking about non-Drupal developers who may end up poking around the modules, could we expand a bit on what this does and why it's useful in the description?

docroot/modules/custom/prisoner_hub_prison_context/tests/src/ExistingSite/PrisonContextTest.php
```diff
@@ -0,0 +1,47 @@
...
+    $route_builder->rebuildIfNeeded();
+  }
+
+  /**
+   * Tests that prison context jsonapi routes have been setup.
+   */
+  public function testRoutes() {
+    /* @var \Drupal\Core\Routing\RouteProviderInterface $route_provider */
+    $route_provider = $this->container->get('router.route_provider');
+    $this->assertGreaterThan(0, $route_provider->getRoutesByPattern('/jsonapi/prison/{prison}/node/article')->count(), 'There is a least one prison context jsonapi route.');
```
spikeheap, line 44: In the ADR we used the format `prisons/{prison}`, but here it's singular. `prisons/` feels like it fits better with the JSON:API naming scheme. Is this way easier/neater?

spikeheap (approved):
Looks good :+1:. I can't say I understand all of it, so it'd be great to get a bit of a walkthrough sometime, but it all looks sensible :). I've added a few minor questions/comments inline.

Minor nit: a lot of the documentation that's been added assumes a working knowledge of Drupal. I wonder if these could be expanded to be more useful for engineers new to Drupal and the codebase?

docroot/modules/custom/prisoner_hub_entity_access/src/EventSubscriber/QueryAccessSubscriber.php
leonk: Yeah it is still outstanding.  Not sure if we want to apply the rule to taxonomy terms as well?  Currently we don't have the prison and prison category field on any taxonomy terms (other than the prison terms itself), so applying the rule wouldn't actually do anything yet.  But if we did, and then added those fields, it would start to work in the same way as it does for content.

docroot/modules/custom/prisoner_hub_entity_access/src/EventSubscriber/QueryAccessSubscriber.php
leonk: So that we can have the flexibility to have some content types _without_ the prison category fields.  We first get all the bundles (i.e. content types) that have the field enabled, and then do we a `OR NOT IN` on those bundles.  I.e. we are only applying the rule to bundles that have the field, on those without we always show the content.

This was the only way I could think of to represent this in a query (as the query at this point could span over multiple bundles).  
Maybe it's worth I add some further documentation on this, as it's probably not the most straightforward thing to grasp?

docroot/modules/custom/prisoner_hub_entity_access/tests/src/ExistingSite/PrisonerHubQueryAccessTest.php
leonk: So..... I was wondering what the requirement was around this, whether the prison field should _override_ the prison category field (i.e. if a prison is set, then ignore anything from the prison category field).  
Or whether they should be independent (i.e. show content if either it matches the prison or prison category field).  
I opted for the latter, which I think is a simpler solution, and means that we are more likely to be showing content rather than hiding it.
I also feel it is more easier to understand to the CMS user.  Otherwise they might select a prison category, expecting to see their content appear, and not notice that there was a prison selected, and suspect something was broken when they couldn't see their content.

But happy to discuss this further.

docroot/modules/custom/prisoner_hub_prison_context/prisoner_hub_prison_context.info.yml
leonk: Yeah good point.  This description was created before I really knew how we were going to implement it.  I can definitely be a bit more specific about it now.

docroot/modules/custom/prisoner_hub_prison_context/tests/src/ExistingSite/PrisonContextTest.php
leonk: Hmmm, another good point.
I guess I was sticking to how Drupal names other resources in JSON:API.  I.e. it's `/node/article` not `/nodes/articles`.
But should it be different here?  As we are talking about a relationship?

leonk:
I've updated some of the documentation to give some more context and hopefully make it easier to understand for non Drupal peeps.

docroot/modules/custom/prisoner_hub_entity_access/src/EventSubscriber/QueryAccessSubscriber.php
leonk: Added some inline comments in the code for this.

docroot/modules/custom/prisoner_hub_entity_access/src/EventSubscriber/QueryAccessSubscriber.php
spikeheap: :+1: assuming it doesn't simplify this code, we should probably add that if/when we update the taxonomy terms 🙌

docroot/modules/custom/prisoner_hub_prison_context/tests/src/ExistingSite/PrisonContextTest.php
spikeheap: :+1: makes sense. I'm ambivalent (as long as the ADR matches the implementation 😉)

leonk: Have just updated the ADR

PR #171, opened 2021-03-16
Installing drush global launcher.

leonk:
### Context

No this is a very small change that doesn't warrant a card.

### Intent

Currently to run drush you need to type in `vendor/bin/drush` which is obviously way too long and causes hours of wasted time and effort.

This PR installs https://github.com/drush-ops/drush-launcher which allows drush to be run from anywhere, and just by typing `drush`.

### Considerations

### Checklist

- [ ] Tested in Development

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -81,6 +81,12 @@ RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" \
   && php composer-setup.php --install-dir=/bin --filename=composer --version=1.10.16 \
   && php -r "unlink('composer-setup.php');"
 
+# Install the global drush launcher
+RUN curl -O https://github.com/drush-ops/drush-launcher/releases/latest/download/drush.phar \
+  && chmod +x drush.phar \
+  && mv drush.phar /usr/local/bin/drush
+
+
 # Set Timezone
 RUN echo "date.timezone = Europe/London" > /usr/local/etc/php/conf.d/timezone_set.ini
 # Set no memory limit (for PHP running as cli only).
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -81,6 +81,12 @@ RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" \
   && php composer-setup.php --install-dir=/bin --filename=composer --version=1.10.16 \
   && php -r "unlink('composer-setup.php');"
 
+# Install the global drush launcher
+RUN curl -OL https://github.com/drush-ops/drush-launcher/releases/latest/download/drush.phar \
```
spikeheap, line 85: Could we pin this to a specific version rather than `latest`? Pinning versions increases our maintenance overhead, but means rebuilding an image should always yield the same results, even if a newer version of drush is released. We saw a problem not so long ago when Composer had a major update with a breaking change that broke our build. That led us to pin Composer up on L81

spikeheap:
Could we achieve the same result by adding `./vendor/bin` to `$PATH`? That way we'd be able to manage the dependency using Composer, but still have it available without the full path in the shell.

leonk:
> Could we achieve the same result by adding `./vendor/bin` to `$PATH`? That way we'd be able to manage the dependency using Composer, but still have it available without the full path in the shell.

Yeah that was the original solution I was going for.  Only real advantage of using the drush launcher is that you can run `drush` from anywhere on the system.  
It's fairly standard practise to use the drush launcher (it used to be part of the drush itself, but they split it out a while back).  
But they list the solution you mentioned in their (alternative)[https://github.com/drush-ops/drush-launcher#alternatives] so perhaps better we do that so we don't add in an additional dependency.

leonk:
Closing in favour of https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/172

PR #189, opened 2021-04-07
Small permission fixes

leonk:
### Context

No these changes are minor and don't warrant a card.

### Intent

Two small changes:

Currently the admin toolbar gives you links to pages where your user account doesn't have any access to the sublinks.
E.g., here I am logged in as a "Studio admin", and I can go to the "Configuration" page even though I can't access anything inside it.
Screenshot 2021-04-07 at 11 47 50
Enabling the `admin_toolbar_links_access_filter` module (which is part of the admin_toolbar module) fixes this issue.

Adding in permissions for local content managers to bulk publish and unpublish for local content managers (as these were removed when we switched to using the views_bulk_operation module to allow bulk field updating).

### Considerations

Local content managers will see a checkbox next to every item of content, including those they don't have access to publish/unpublish.  If they try to bulk update these, they get an "Access Denied" error.
Screenshot 2021-04-07 at 11 55 25

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #191, opened 2021-04-23
36 Rename date field improve description

leonk:
### Context

https://trello.com/c/h4QLGZyy/36-rename-and-improve-date-field-series-sorting-only

### Intent

This PR adds the following functionality:
- Introduces two new options to the `field_sort_by` on the series Taxonomy term. _"Release date (newest first)"_ and _"Release date (oldest first)"_
- Renames `field_moj_date` to `field_release_date`.  This is done by creating a new field, hiding the existing field, and copying the contents across.  The previous field will then be removed on a subsequent deployment.
- Use Drupal's javascript form states API, to show/hide the season+episode number and release date fields depending on what series has been selected (and what `field_sort_by` value has been assigned to that series).
- Improve field labels and descriptions as per work done with Sophie.  The `field_group` module has been added to group fields and make the UX a bit nicer.
- Added date fields and sorting to the "Change episode order" admin page.
- Added Javascript tests with Webdriver.

The screenshot below shows the field labels and descriptions (although doesn't demonstrate the show/hide functionality).
image

### Considerations

There are dependencies on specific configuration (e.g. field names) that are not explicitly defined as dependencies in the `prisoner_hub_taxonomy_sorting` module.  To make the module not have any dependencies on the configuration, we would need to create an abstraction (most likely with even more configuration), which I don't think is worth it.

There is some styling tweaks made in  `_prisoner_hub_taxonomy_sorting_fix_field_group_descriptions()` (found in the `prisoner_hub_taxonomy_sorting.module` file).  These would ideally be in a theme, but we currently do not have our own theme to run them from.  I believe at somepoint we will switch to our own theme, possibly either when we totally switch everything over to GDS, or when we upgrade to Drupal 9 (which has a new admin theme).  So we should move this code at this point.

### Checklist

- [ ] Tested in Development

### DIFF
--- a/config/sync/core.entity_form_display.node.moj_pdf_item.default.yml
+++ b/config/sync/core.entity_form_display.node.moj_pdf_item.default.yml
@@ -15,16 +15,62 @@ dependencies:
     - field.field.node.moj_pdf_item.field_moj_thumbnail_image
     - field.field.node.moj_pdf_item.field_moj_top_level_categories
     - field.field.node.moj_pdf_item.field_prison_categories
+    - field.field.node.moj_pdf_item.field_release_date
     - image.style.thumbnail
     - node.type.moj_pdf_item
   module:
     - datetime
+    - field_group
     - file
     - image
     - path
     - scheduler
     - select2
     - text
+third_party_settings:
+  field_group:
+    group_series:
+      children:
+        - field_moj_series
+        - group_season_and_episode_number
+        - group_release_date
+      parent_name: ''
+      weight: 10
+      format_type: fieldset
+      region: content
+      format_settings:
+        description: 'Select which series this episode belongs to. Then, select either a season and episode number <strong>or</strong> a date.'
+        required_fields: true
+        id: ''
+        classes: ''
+      label: 'How does this episode fit into a Series?'
+    group_season_and_episode_number:
+      children:
+        - field_moj_season
+        - field_moj_episode
+      parent_name: group_series
+      weight: 18
+      format_type: fieldset
+      region: content
+      format_settings:
+        description: 'Enter a season and episode number.'
+        required_fields: true
+        id: ''
+        classes: ''
+      label: 'What is the season and episode number?'
+    group_release_date:
+      children:
+        - field_release_date
+      parent_name: group_series
+      weight: 19
+      format_type: fieldset
+      region: content
+      format_settings:
+        description: "<p>Enter the date of the original service or show, or when the content was first aired.</p>\r\n<p>This date will affect the order of the content displayed within the series.</p>"
+        required_fields: true
+        id: ''
+        classes: ''
+      label: 'Which date would you like to assign to this content?'
 _core:
   default_config_hash: Oo1-wQanlhvJ6-TSJzG1NYGLiNQTTRYKYbcOAPOE09g
 id: node.moj_pdf_item.default
@@ -45,12 +91,6 @@ content:
     third_party_settings: {  }
     type: boolean_checkbox
     region: content
-  field_moj_date:
-    weight: 14
-    settings: {  }
-    third_party_settings: {  }
-    type: datetime_default
-    region: content
   field_moj_description:
     weight: 6
     settings:
@@ -62,7 +102,7 @@ content:
     type: text_textarea_with_summary
     region: content
   field_moj_episode:
-    weight: 12
+    weight: 13
     settings:
       placeholder: ''
     third_party_settings: {  }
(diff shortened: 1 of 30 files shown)


config/sync/core.entity_form_display.node.moj_pdf_item.default.yml
```diff
@@ -15,16 +15,62 @@ dependencies:
     - field.field.node.moj_pdf_item.field_moj_thumbnail_image
     - field.field.node.moj_pdf_item.field_moj_top_level_categories
     - field.field.node.moj_pdf_item.field_prison_categories
+    - field.field.node.moj_pdf_item.field_release_date
```
andrewrlee, line 18: why is this in a file called pdf_item?

config/sync/core.entity_form_display.node.moj_pdf_item.default.yml
```diff
@@ -142,25 +188,25 @@ content:
     type: boolean_checkbox
     settings:
       display_label: true
-    weight: 15
+    weight: 13
```
andrewrlee, line 191: is weight like an ordering on the page kind of thing?

config/sync/core.entity_form_display.node.moj_pdf_item.default.yml
leonk: There is a content type called "PDF item" which this file refers to.  These can be part of a series on the Hub (along with audio items, video items, and basic pages). 
We did take a look at removing PDF items from series, but there is already quite a bit of content that is created as pdf's as part of a series.

config/sync/core.entity_form_display.node.moj_pdf_item.default.yml
leonk: Yep the weight controls the order of the fields on the form.  

All of these types of yml changes are generated by Drupal (i.e. not directly modified in code).  I've made changes to through the Drupal UI, and then ran the config export.  So this means you often get small changes like this that weren't entirely intentional (but aren't an issue other than it makes PR's difficult to review).

docker-compose.circleci.yml
```diff
@@ -23,6 +23,12 @@ services:
       - 11001:8080
     networks:
       - circleci-prisoner_content_hub
+  chrome:
+    image: previousnext/chrome-headless:65
```
andrewrlee, line 27: bit of a random org to be installing docker images from. Have you tried any of the circle images? https://circleci.com/developer/images/image/cimg/node#browsers

These tend to be much faster as massively cached but also come from a trusted source

andrewrlee: or maybe https://circleci.com/developer/orbs/orb/circleci/browser-tools?

docroot/modules/custom/prisoner_hub_taxonomy_sorting/prisoner_hub_taxonomy_sorting.deploy.php
```diff
@@ -26,3 +27,14 @@ function prisoner_hub_taxonomy_sorting_deploy_set_field_default_values() {
...
 }
+
+
+/**
+ * Copy values from field_moj_date to field_release_date.
+ */
+function prisoner_hub_taxonomy_sorting_deploy_copy_moj_date_2() {
+  // @See https://drupal.stackexchange.com/a/250937/4831
+  Drupal::database()->query("INSERT INTO node__field_release_date SELECT * FROM node__field_moj_date;");
+  Drupal::database()->query("INSERT INTO node_revision__field_release_date SELECT * FROM node_revision__field_moj_date;");
```
andrewrlee, line 38: so this is some kind of migration - does this get run automatically?

docker-compose.circleci.yml
leonk: This came from the Drupal test traits library that we're using.  https://gitlab.com/weitzman/drupal-test-traits/-/blob/master/docker-compose.yml#L47
I didn't try out any other images (or work out what this specific image is doing differently).  It might be that the Drupal test traits library has a dependency on that particular image (if this is the case then I guess we should document it).
I could take a further look in another PR?

docroot/modules/custom/prisoner_hub_taxonomy_sorting/prisoner_hub_taxonomy_sorting.deploy.php
leonk: Yep this is a migration from one field to another (which is the only way we can rename the field).

And yes, it gets run automatically as part of the deployment.

config/sync/core.entity_form_display.node.moj_pdf_item.default.yml
andrewrlee: oh right! that makes sense

andrewrlee (approved):
LGTM

PR #190, opened 2021-04-12
49 add ability to cms to display and edit a series order

leonk:
### Context

> Does this issue have a Trello card?

https://trello.com/c/hRFu6ke3/49-add-ability-to-cms-to-display-and-edit-a-series-order

and a the very start of...
https://trello.com/c/ThDYVbQ1/46-add-ability-to-sort-series-by-oldest-first (as the field for this one needs to be setup in order to do the other Trello card).

### Intent

This PR adds a _temporary_ page tab to all Series Taxonomy terms, `Change Episode Order`.  This allows the content within a Series to be updated, in order to obtain the correct sorting on the frontend.

This PR also introduces a new `field_sort_by` on the Series Taxonomy term.  This field does not currently effect the order on the frontend, this will be implemented as part of https://trello.com/c/ThDYVbQ1/46-add-ability-to-sort-series-by-oldest-first

### Checklist

- [ ] This PR contains **only** changes related to the above card
- [ ] Tests have been added/updated to cover the change
- [ ] Documentation has been updated where appropriate
- [ ] Tested in Development

PR #192, opened 2021-04-27
46 add ability to sort series by oldest first

leonk:
### Context

https://trello.com/c/ThDYVbQ1/46-add-ability-to-sort-series-by-oldest-first

### Intent

This PR implements the new sorting methodology for our rest endpoints, which means that the sorting on series will be updated in the frontend.
The previous sorting function has been removed, and instead sorting is applied on the entity query itself (which is more performant, and will fix the issue where series with 40+ episodes have missing content).
The value of the `field_sort_by` (added in https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/190) is now used to determine the sorting method.

### Considerations

No automated tests have been added/updated for this change, (the previous unit tests for these rest endpoints no longer work, and we will be moving to JSON:API in the near future so not much value in fixing and updating these).

### Checklist

- [ ] Tested in Development

### DIFF
--- a/docroot/modules/custom/moj_resources/src/SeriesContentApiClass.php
+++ b/docroot/modules/custom/moj_resources/src/SeriesContentApiClass.php
@@ -108,7 +108,7 @@ class SeriesContentApiClass
       $result = [];
       $result["episode_id"] = $episodeId;
       $result["last_updated"] = $curr->changed->value;
-      $result["date"] = $curr->field_moj_date->value;
+      $result["date"] = $curr->field_release_date->value;
       $result["content_type"] = $curr->type->target_id;
       $result["title"] = $curr->title->value;
       $result["id"] = $curr->nid->value;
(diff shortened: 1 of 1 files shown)


docroot/modules/custom/moj_resources/src/SeriesContentApiClass.php
```diff
@@ -225,9 +185,41 @@ private function getSeriesContentNodeIds($seriesId, $number, $offset, $prison)
...
       ->accessCheck(false);
-
     $results->condition('field_moj_series', $seriesId);
 
+    $seriesTerm = Term::load($seriesId);
+    $sortByFieldValue = $seriesTerm->get('field_sort_by')->getValue();
+    $sortByFieldValue = empty($sortByFieldValue) ? NULL : $sortByFieldValue[0]['value'];
+
+    $sortFields = [];
+    $sortDirection = '';
```
spikeheap, line 195: Are these two lines needed? It looks like the switch statement will always set a value, so these lines may be redundant.

spikeheap (dismissed):
Looks sensible 👍

docroot/modules/custom/moj_resources/src/SeriesContentApiClass.php
leonk: Ah yeah that was more from a general habit of setting up variables with default values before setting them.  But you're right as there is the `default` in the switch statement, these aren't needed.  I've removed them.

spikeheap (approved):

PR #196, opened 2021-05-07
Prep work for implement prison category rule on topics page

leonk:
### Context

https://trello.com/c/X0y3aNeI/113-implement-prison-category-rule-on-topics-page

### Intent

This is some prep work for implementing JSON:API.  To fix an issue discovered by @andrewrlee where the homepage `featured_articles` content type was not working correctly with the prison category filtering.  It was instead returning content for all prisons.

To fix this, the prison category field has been added to the `featured_articles` content type.  Previously it just had the prisons field (but not categories).  The prison category filtering doesn't support content types with _just_ the prisons field (rather than adding this support, it's easier to just enable the field so that this content type is the same as all the others).

I then modified the tests to ensure every content type with the prison field was tested.  Which made me discover that the `moj_hub_item` and `article` content types were failing.  Looking into this, I can see that both content type are not being used (all of the content is unpublished).  So both content types have been removed as part of this PR (as this is easier/better than fixing a redundant content type).

### Considerations

I checked with Lauren and Jo about the removal of the `moj_hub_item` and `article` unused content types, both confirmed they were OK with the removal.

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM!

PR #200, opened 2021-05-12
113 implement prison category rule on topics page 2

leonk:
### Context

https://trello.com/c/X0y3aNeI/113-implement-prison-category-rule-on-topics-page

### Intent

Introduces a new jsonapi resource `/jsonapi/topics`.
This will also be available when using the prison in the url.  E.g. 
`/jsonapi/prison/berwyn/topics`

### Considerations

As part of converting the /topics page to JSON:API, I've started the process of removing the `landing_page` content type (it won't be removed just yet, but the /topics JSON:API response will only return the taxonomy terms, and not the landing page node type).
To allow the urls to the existing landing pages to be generated, a new field has been added `field_legacy_landing_page`, which contains the reference back from category to landing page (this is also kept in sync by using the "Corresponding Entity References" module.  Once we have converted the category pages (and removed the landing_page content type), we can remove this field.

Prison and Prison category fields have been added to Categories and Secondary tags.  For categories the values of their corresponding landing pages have been copied across.  For tags, all prison categories have been selected (for every tag).

### Checklist

- [ ] Tested in Development


docroot/modules/custom/prisoner_hub_jsonapi_resources/prisoner_hub_jsonapi_resources.routing.yml
```diff
@@ -0,0 +1,6 @@
+prisoner_hub_jsonapi_resources.topics:
+  path: '/%jsonapi%/topics'
```
leonk, line 2: Hmm I think I might change this.  As it's not strictly valid in the JSON:API spec.  Firstly it should be "topic" instead of "topics".  And if we have a resource called "topic", then to follow the spec, each individual resource should be available underneath this.  I.e. `/jsonapi/topic/{{uuid}}`.

The alternative approach is to enable the endpoint `jsonapi/taxonomy_term` (which can be done via the https://www.drupal.org/project/jsonapi_cross_bundles) module).  You could then use a `?filter` parameter to limit the items to be just categories and tags.  Unfortunately, the filter looks a little fiddly due to some weird quirk in Drupal, and having to use an `IN`.  The final url would look a little bit like this:
`jsonapi/prison/berwyn/taxonomy_term?filter[vid-filter][path]=type.meta.drupal_internal__target_id&filter[vid-filter][operator]=IN&filter[vid-filter][value][1]=moj_categories&filter[vid-filter][value][2]=tags`

andrewrlee (approved):
LGTM

andrewrlee (approved):
LGTM

andrewrlee (approved):
LGTM

leonk:
Closed in favour of https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/201

PR #206, opened 2021-05-21
Moving patches to local files

leonk:
### Context

No this is a small change that doesn't require a card.

### Intent

Patches have been added that download directly from drupal.org and gitlab.com, these have been moved to local files.  This avoids a dependency in our builds on external services being available.  And also reduces the risks of supply-chain (package manager attacks).  Although files hosted on drupal.org should be permanent (i.e. once uploaded they can't be edited or removed), relying on this functionality is a security risk.

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #209, opened 2021-05-27
Update drupal core 8.9.16

leonk:
### Context

Nope, it's a small change.

### Intent

Bump Drupal version number up to 8.9.16, to include security fix for ckeditor.
We _should_ not be particularly vulnerable, as ckeditor is only used by people with a CMS account.

### Considerations

When updating Drupal core, I noticed that a contrib module `views_ef_fieldset` that was recently added, was applying a patch to drupal core.  This was being downloaded directly from drupal.org.

Since https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/206 we made a decision to only patch from local files.  
The patch from the module did not add any functionality, and was only used for the modules internal testing.  So it has been removed (by placing it in `patches-ignore`).

### Checklist

- [ ] Tested in Development

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -122,6 +122,13 @@
             "drupal/views_ef_fieldset": {
                 "Issue #3173822: Exposed operators are not included in fieldsets": "patches/3173822-views_ef_fieldset-operator_issue-3.patch"
             }
+        },
+        "patches-ignore": {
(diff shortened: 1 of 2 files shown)


andrewrlee (approved):
LGTM

PR #221, opened 2021-07-07
180 automatically re size images to increase frontend performance

leonk:
### Context

https://trello.com/c/2uN9A0P5/180-automatically-re-size-images-to-increase-frontend-performance

### Intent

Use Drupal's image resizing functionality to resize images.

Two new "image styles" are introduced in this PR.  These are essentially presets (which can be updated at anypoint in time).
I based these sizes on what is currently used on the frontend.

- `tile_small` (296x176) - This is the most commonly used image size, used on most taxonomy pages and the homepage.
- `tile_large` (600×345) - Larger images are currently used in three separate places, all with slightly different sizes.  The homepage (600x300), category pages (585x345), and at the top of a series page (450x300).  These sizes are all quite similar so it doesn't make sense to have separate image styles for each one (it would result in 3x images being created).   So instead I took the largest of each dimension, and that became the size.

I've also updated the patches for the flysystem_s3 module to support image styles, in particular the patch for the pre-signed urls.  This has been posted back to the original issue queue, see https://www.drupal.org/project/flysystem_s3/issues/2772847#comment-14156598

The jsonapi_image_styles module has been installed, so that image style urls can be retrieved via JSON:API.
To do this, you must use the `?include` parameter.  E.g. http://localhost:11001/jsonapi/node/moj_video_item?include=field_moj_thumbnail_image

### Checklist

- [ ] Tested in Development


leonk:
I've spotted an issue when viewing a page with lots of images for the first time (e.g. a series page with 40 images).  Drupal will need to resize each one, and it hasn't done this in time when the user loads the page.  Resulting in broken images.

I'm going to look into a fix for this possibly using https://www.drupal.org/project/image_style_warmer

leonk:
Re-opened PR with adding `image_style_warmer` module enabled.
This module will allow us to generate resized images for all existing files.  This can be done by running `drush image-style-warmer:warm-up`.  Note the process will take around 4-5 hours to run.

leonk:
There's still a bit of work to do as the k8ns job to generate the images was failing.  So closing this for now while that is fixed.

PR #225, opened 2021-07-15
180 re size images deploy first

leonk:
### Context

Please see original PR https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/221 (which has now been split into three PRs)

### Intent

This is the first part of a series of deployments for this work.

These changes must be deployed first, before we can run the k8ns job to generate the image styles.

### Considerations

The bulk of the changes here are in the patch to the `flysystem_s3` module, which is in the `patches/flysystem_2772847-54-private-files-image-styles.patch` file.  This has been taken from the latest patch found on http://drupal.org/node/2772847 
We were previously using an older version of the patch, which didn't support image styles.

The latest version was created by me! https://www.drupal.org/project/flysystem_s3/issues/2772847#comment-14156598
But it only contains a few small tweaks from before, so should be safe to use (i.e. has been tested by others in the community).  Unfortunately, there are no tests specifically for private buckets in the module (I imagine it's probably quite a difficult thing to test against).

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #226, opened 2021-07-15
180 resize images deploy second

leonk:
This PR should be deployed after https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/225

It introduces a helm task to generate the image styles.  We need to have the updates provided by the previous PR on live before we can run this task.


andrewrlee (approved):
LGTM

PR #228, opened 2021-07-28
Fix for Drupals render caching showing expired S3 urls.

leonk:
### Context

https://trello.com/c/2uN9A0P5/180-automatically-re-size-images-to-increase-frontend-performance

This issue was spotted when testing https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/225
Drupal would cache the S3 url for longer than the signature was valid for, eventually resulting in invalid urls.  

### Intent

This PR updates the patch to the flysystem_s3 module to add in the correct `max-age` based on the expires setting for the signature.  See https://www.drupal.org/project/flysystem_s3/issues/2772847#comment-14175477 for more info.

### Considerations

We still need to be careful with caching for S3 urls, if we enable any additional cache's in the future (e.g. a reverse proxy cache) then we will need to ensure the TTL's are set correctly.
Note this isn't a new issue, and is not directly caused by the previous PR.  Currently all caching is disabled, but with the resizing of images + JSON:API, we will be enabling _some_ of Drupal's caches, so it has now become an issue.

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #229, opened 2021-07-28
Updating to drupal core 9.1.11 to include security fix

leonk:
### Context

### Intent

Bumps Drupal core version to 9.1.11 to include security fix https://www.drupal.org/sa-core-2021-004

We are unlikely to be vulnerable to the security issue, as we do not have any custom or contributed code using the Archive_Tar library.  But we should update anyway to be sure.

### Considerations

na

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #230, opened 2021-08-05
233 switch to jsonapi for basic pages

leonk:
### Context

https://trello.com/c/fh1sP4L6/233-switch-to-jsonapi-for-basic-pages

### Intent

This PR adds the following functionality in preparation for FE work for switching to JSON:API for basic pages.

- Enables the [decoupled_router](https://www.drupal.org/project/decoupled_router) module.  This provides an API endpoint for looking up paths.  From this the appropriate template can be selected in the frontend.  See example request below:
Screenshot 2021-08-05 at 11 02 21
- Installs the [pathauto](https://www.drupal.org/project/pathauto) module, and sets up url patterns the same as currently being used on the FE.  I.e. `content/{id}` and `tags/{id}`.  Drupal will now store these paths, which we can one day make configurable via the CMS.  Although for now the option to modify a path is hidden from all account levels except for Administrators.

### Considerations

Currently the url to lookup a path is `/router/translate-path?path=content/123`, however this doesn't take into account the current prison.  This means two things:
- The jsonapi url returned in the response will not include the `prison/wayland` part.
- A check to see that the content is available in the current prison won't be performed (although this will happen on the second JSON:API request, so the user still won't be able to get to the content.  Also, any path being looked up this way would have come from a previous JSON:API request, e.g. the homepage or topics page, so the access check would have already happened there).

We could either:
a) Modify the endpoint to be something like `router/prison/wayland/translate-path?path=content/123`
b) Generate the JSON:API url in JS, using something like: `jsonapi.entryPoint + '/prison/' + establishmentName + '/' + entity.type + '/' + entity.bundle + '/' + entity.uuid`

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

leonk:
We decided to go with option a for the decoupled router path, this has been implemented in https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/231

PR #232, opened 2021-08-11
233 add extra entity access check

leonk:
### Context

This is work related to preparing for:

https://trello.com/c/fh1sP4L6/233-switch-to-jsonapi-for-basic-pages 

### Intent

The previous PR https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/231 modified the `/router/translate-path` callback to use the current prison, i.e. `/router/prison/wayland/translate-path`.
However, the access check (for whether content was assigned to the current prison/category) is not actually run on this request. (This means you would still get a 200 response, and then a 403 on the subsequent JSON:API request).

This PR introduces the use of `hook_entity_access()`, which could be considered a _last resort_ access hook (i.e. if the other API's we are using end up being bypassed in someway, this hook will _always_ be respected no matter what).

There is also a bit of a cleanup/refactor of the `prisoner_hub_entity_access` module.  Including an update to the documentation in the `README.txt` file, explaining the three different entity access APIs being used.

Example of a router request for content that is not available on the current prison.

Screenshot 2021-08-11 at 12 16 19

### Considerations

### Checklist

- [ ] Tested in Development


davidthomas22 (approved):
Conducted review with Leon - LGTM

PR #233, opened 2021-08-11
Create paths for all nodes, including unpublished.

leonk:
### Context

Fix relating to previous PR As these could be linked through somewhere on the site currently.

### Intent

Previously unpublished nodes were not given paths.  As we know that we have some unpublished content appearing on the site, we need to make sure they also have a path, so that the `router/translate-path` request can find it.

### Considerations

### Checklist

- [ ] Tested in Development


davidthomas22 (approved):
LGTM

PR #234, opened 2021-08-12
251 Fix missing images on homepage

leonk:
### Context

https://trello.com/c/SyGucW4v/251-missing-images-on-homepage

### Intent

This PR uninstalls the dynamic_page_cache module, as this was causing an issue where it was _re-using_ the jsonapi cache, and setting a new max-age.
So if the jsonapi cache had been generated at somepoint in the past, say 20 mins ago.  Any image urls generated in that cache would only now be valid for 40 mins.  But the dynamic_page_cache would set the max-age to be 60 mins, meaning that in the last 20 mins the image urls would be invalid.

I have also updated the patch for the flysystem_s3 module, to remove parts that were no longer needed.  

See https://www.drupal.org/project/flysystem_s3/issues/2772847#comment-14188928 for more info.

### Considerations

Should probably write some documentation about the S3 url signatures.

### Checklist

- [ ] Tested in Development


davidthomas22 (approved):
LGTM

PR #236, opened 2021-08-16
Fix issue where image urls inside the text editor were expiring.

leonk:
### Context

This fix is related to https://trello.com/c/fh1sP4L6/233-switch-to-jsonapi-for-basic-pages

### Intent

This fixes the issue of files uploaded through the text editor getting expired urls.
The cache metadata of the file was not being properly passed onto the containing content.  So whilst we are setting a max-age for the file based on the url signature expiration, this was not being used when the file was uploaded inside a text editor field.

This is a bug in Drupal, I opened an issue for this: https://www.drupal.org/project/drupal/issues/3228329

### Considerations

As this is a patch to Drupal core, we should monitor the above issue to see whether this is an acceptable solution or not.
It may also be worth writing tests for this, to ensure that the fix gets committed back in.

### Checklist

- [ ] Tested in Development

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -122,7 +122,8 @@
             },
             "drupal/core": {
                 "Issue #1149078 - States API doesn't work with multiple select fields": "patches/drupal_states-multiselect-1149078-109.patch",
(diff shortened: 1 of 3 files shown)


andrewrlee (approved):
LGTM

PR #238, opened 2021-08-19
254 switch to jsonapi for series pages

leonk:
### Context

https://trello.com/c/EyX2RFzY/254-switch-to-jsonapi-for-series-pages

### Intent

This PR adds a new `series_sort_value` field, which stores the calculated value for sorting of an episode within a series.  It is based on what type of sorting has been selected for the series. 

If sorting by season+episode number is selected, the field will be a concatenation of season and episode (with episode number padded to three digits).  E.g. season 1, episode 13 will be 1013.  Season 15 episode 2 will be 15002.

If sorting by release date is selected, then the field will be the timestamp of the `field_release_date`.  E.g. 1629388758

If sorting by DESC is selected (i.e. Newest first), then the values are inverted to negative values.  
E.g. -1013, or -1629388758 (from the examples above).

The goal is to have one method for sorting via JSON:API.  So you would always use the same field and direction.  With this PR, you can use something like:
`/jsonapi/node?filter[field_moj_series.meta.drupal_internal__tid]=123&sort=series_sort_value`
And this will work for all types of sorting selected on the series.

### Considerations

### Checklist

- [ ] Tested in Development

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ commands:
           command: |
             VERSION_TO_DEPLOY=$(cat /tmp/build-info/version-to-deploy.txt)
             helm upgrade ${HELM_BACKEND_RELEASE_NAME} ~/git/helm_deploy/prisoner-content-hub-backend \
-              --install --wait --force --reset-values --timeout 360s \
+              --install --wait --force --reset-values --timeout 1200s \
               --namespace=${KUBE_NAMESPACE} \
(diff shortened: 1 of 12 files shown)


leonk:
This PR needs to stay in draft until https://trello.com/c/I3GkyBuT/261-only-allow-content-to-be-linked-to-one-series has been finished.

andrewrlee (approved):
LGTM

PR #244, opened 2021-09-01
Adding description field as an exposed filter. …

leonk:
### Context

No this is a small change that doesn't warrant a card.

### Intent

This PR adds the description field as a filter to the /admin/content page, which can be used to search for embedded urls.

We've been seeing lots of 404s appearing in the logs recently.  Some of which are from embedded links inside pages.  Adding this filter will allow the content team to search through the site for places where they should be updated.

Screenshot 2021-09-01 at 16 06 43

### Considerations

### Checklist

- [ ] Tested in Development

### DIFF
--- a/config/sync/views.view.content.yml
+++ b/config/sync/views.view.content.yml
@@ -1430,6 +1430,50 @@ display:
           hierarchy: false
           error_message: true
           plugin_id: taxonomy_index_tid
+        field_moj_description_value:
+          id: field_moj_description_value
+          table: node__field_moj_description
+          field: field_moj_description_value
+          relationship: none
+          group_type: group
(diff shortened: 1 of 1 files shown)


andrewrlee (approved):
LGTM

leonk:
Small update to limit the available operators, so that only ones that make sense (to be used with) this field are included.

Before we had things like "Is equal to", which would be very difficult to use, as all description fields start with an opening `<p>` tag (so the user would need to ensure they search for _exactly_ the right markup).  
Here are the new set of operators:
Screenshot 2021-09-03 at 11 21 36

andrewrlee (approved):
LGTM

PR #246, opened 2021-09-06
Fix deployment health check issue

leonk:
### Context

The deployments on main branch are currently failing, the k8ns fails and the pods are restarted.
Because the Drupal site gets put into maintenance mode during the deployment, and this disables the health check.

### Intent
This PR does two things:
1) Installs the maintenance_exempt module, and configures it to allow the health check, JSONAPI, and REST api urls to still be accessible when the site is in maintenance mode.
2) Moves the problematic update hook to a deploy hook, which does not operate in maintenance mode.

### Considerations

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #247, opened 2021-09-09
Fix issue with missing sort values

leonk:
### Context

No, this is an issue reported on Slack:

https://mojdt.slack.com/archives/C9EKLJ1QE/p1631176829002200

### Intent

The last deployment introduced a change where when content is saved, a sort value is calculated.  This value is based on the season+episode number fields *or* the release date field (depending on what series had been selected).

The error was caused when the value was empty, and the content was first being added.
This is because of a warning message printed out, which included a link to the content.  At this point, there was no url for the content, so there was a PHP error.

Also, to avoid user submitting empty values, I've made the season+episode number fields required.
The release date field I couldn't make required (due to an issue in drupal core, which is detailed in the code comments).  But instead the release date field is given a default value of today's date.  So a user would need to unset this, in order to submit an empty value.

### Considerations

### Checklist

- [ ] Tested in Development

### DIFF
--- a/config/sync/field.field.node.moj_pdf_item.field_release_date.yml
+++ b/config/sync/field.field.node.moj_pdf_item.field_release_date.yml
@@ -15,7 +15,10 @@ label: Date
 description: ''
 required: false
 translatable: true
-default_value: {  }
+default_value:
+  -
+    default_date_type: now
+    default_date: now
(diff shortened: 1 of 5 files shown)


andrewrlee (approved):
LGTM

PR #250, opened 2021-09-10
Add delete file logging

leonk:
Context
Does this issue have a Trello card?

Intent
This PR adds logging to when files are deleted from S3. This logging will only apply to files that are deleted through Drupal.

Previously it was thought that files were never deleted in Drupal. But since enabling S3 file logging ministryofjustice/cloud-platform-environments#5497, we have discovered that Drupal is sometimes deletes files. We disabled logging ministryofjustice/cloud-platform-environments#5551 as it was no longer required, and will use the logging provided by this PR to further investigate the issue.

Considerations
Is there any additional information that would help when reviewing this PR?

Are there any steps required when merging/deploying this PR?

Checklist
 This PR contains only changes related to the above card
 Tests have been added/updated to cover the change
 Documentation has been updated where appropriate
 Tested in Development


andrewrlee (approved):
LGTM

PR #253, opened 2021-09-16
Removing old elasticsearch index, 

leonk:
that has not been used since we switched to JSON:API for search.

### Context

No this is a very small change.

### Intent

When we moved to using JSON:API for search, a new index was created.  We forgot to remove the old one.

This PR removes the old index.  This should help with performance as currently content is indexed twice.

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #258, opened 2021-09-30
Adding sentry as an output for monologger.

leonk:
### Context

https://trello.com/c/xL7P4r7Z/274-fix-drupal-and-sentry-integration

### Intent

This should hopefully fix the Sentry integration, there appears to be an issue integrating with the monologger module (which was enabled to allow logging in cloud platform).

### Considerations

If this doesn't work, the next step would be to update the sentry module and client.

### Checklist

- [ ] Tested in Development


davidthomas22 (approved):
LGTM

PR #283, opened 2021-11-04
Enable drupal page caching

leonk:
### Context

https://trello.com/c/PxWAxREE/392-enable-drupal-page-caching

### Intent

This PR enables page caching in Drupal, this should have a significant performance impact on the site.

The issue that prevented us from previously enabling page caching, was that our S3 urls were set to expire in 1 hour.  Drupal's page cache module was not able to detect this, and would set a permanent expiration.  

This PR fixes this issue by using the patch from https://www.drupal.org/project/drupal/issues/2352009#comment-14064154

Also, the expiration of S3 urls has been updated to be an absolute time: the next day at 3am.
Using an absolute time (as opposed to relative) will resolve issues where cache's are re-used within Drupal.  For example, if an image url has already been generated on another page, the cached url can be re-used, but with a new max-age set.  This is the issue we're having in: https://trello.com/c/vvbPAF4Z/340-fix-issue-with-broken-expired-images-appearing-on-the-homepage

### Considerations

The biggest consideration is incorporating the patch, as is it's not been merged into Drupal (and probably unlikely to be merged in soon).   So isn't fully tested (and does not contain it's own automated tests).

However, I believe that the reason it has not been merged, is because _most_ large sites use Drupal with an external cache (Varnish or a CDN).  The internal page cache module is mostly intended for small sites.  Going over the comments from the [issue](https://www.drupal.org/project/drupal/issues/2352009), it seems as thought the maintainers are unwilling to merge the patch in, as it could result in pages being un-cacheable across lots of small Drupal sites (which could potentially break the internet).  

As we are aware of what we're doing (sort of), I believe we should be OK to use this patch (although this should be documented in Confluence).
The other thing to note is that most smaller Drupal sites will be using Drupal as their frontend, so the caching is much more complicated with many components making up a page.  For us, as we are using JSON:API, it's a lot simpler.  The only max-age we need to deal with is that of S3 url signatures.

An alternative solution is using a contrib module [cache_control_override](https://www.drupal.org/project/cache_control_override), but that also requires a patch: https://www.drupal.org/project/cache_control_override/issues/2962699
I believe it is probably safer to use the Drupal core patch, as it is smaller, and effects the exact piece of code responsible for setting the expiration.

Another consideration is whether there are any issues between using two different servers for setting expirations.  I.e. the server Drupal runs on and S3.  As we are setting a specific timestamp for both S3 and page cache expiration, this should in theory not be an issue, both timestamps will be UTC. 

### Checklist

- [ ] Tested in Development


leonk:
Closed as github got confused about what it was merging in.  Re-opened in https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/290

PR #285, opened 2021-11-05
Revert "Revert "Fix unpublished content being shown""

leonk:
This PR re-instates https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/277

Which was reverted in ministryofjustice/prisoner-content-hub-backend#278

Turns out the original PR was correct, and the missing content on homepages (which caused us to revert the change) was actually caused by unpublished content, so should not be shown.


andrewrlee (approved):
LGTM

PR #289, opened 2021-11-08
Remove old api part 1

leonk:
### Context

https://trello.com/c/SSyiiOXK/269-turn-off-the-old-custom-drupal-api

### Intent

This is the first part in a two step process for removing the old API.

We first need to uninstall the modules related to the old API, once that has been done and then deployed, we can remove the code from the codebase.

Modules uninstalled in this PR:

- moj_resources (this contained all of the old API endpoints)
- moj_healthcheck (this was technically not part of the old API, but was using the Drupal rest module, which we want to uninstall.  So instead a contributed module called [health_check](https://www.drupal.org/project/health_check) is enabled).
- rest (core) module

### Considerations

We should test for any PHP errors relating to the removal of the old API, and be vigilant once this is deployed to prod.

Also, the k8ns health check has now been altered slightly.  The new one just returns a timestamp, without any formatting (previously it would return a JSON response).  I'm assuming all that's required is a valid status code, but we should double check this once it's deployed to dev/staging

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

andrewrlee (approved):
LGTM

PR #295, opened 2021-11-10
Fix the mysterious missing file issue

leonk:
### Context

### Intent

This PR fixes the longstanding missing file issue.

The issue was caused by a bug in the flysystem_s3 module. Where if an upload failed, and then the same filename was re-uploaded, Drupal would delete the file as part of its cleanup operations.

The bug is better described here: https://www.drupal.org/project/flysystem_s3/issues/3248466

This PR adds (yet another) patch to the flysystem_s3 module, where it checks whether the file exists as a record in the Drupal database, and if so, gives it a unique filename.
This way filenames should always be unique, and when Drupal cleans up the files, it doesn't inadvertently delete live files.

This PR also enables JS Sentry error reporting.  This should hopefully capture any additional issues with the file uploads.  As it appears that often file uploads fail, and as they are going directly to S3, we don't know anything about the issue in PHP.

### Considerations

We are now using 5 separate patches on the flysystem_s3 module, and it's evident this is quite a brittle part of our Drupal instance.
If we continue to see issues with the file uploads, we could consider disabling the upload progress bar (CORS), as this currently requires three patches (including the one from this PR) to get it to work.

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #297, opened 2021-11-11
Disable other file fields and submit buttons when uploading.

leonk:
### Context

and
https://trello.com/c/ididuzch/227-prevent-user-from-saving-content-before-a-file-has-finished-uploading

### Intent

Currently multiple files can be uploaded simultaneously, i.e. a video and a thumbnail can be uploaded at the same time.

When this happens, the first file that finishes uploading to S3, has to actually wait for the second file to upload, and this file ends up getting sent directly to Drupal (in addition to S3), which is totally unnecessary.   This issue is described here: https://www.drupal.org/project/flysystem_s3/issues/3248956
Whilst this doesn't actually cause any errors, if both files are uploaded successfully, it does slow things down.  It also means that if one file fails, then it can impact the other.

During investigating the missing file issue, we found that often video files would fail to upload, and get re-uploaded.  (This caused the error where the first failed attempt ended up causing the successful attempt to be deleted, see https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/295 for more details).  This issue has now been fixed, so it's OK for files to fail to upload, but this PR is to stop failed uploads from happening as often.

To fix the issue, we disable all _other_ file fields, as well as submit buttons (as pressing submit during an upload will break the upload). 

Before (notice how the second file gets stuck on 100%)

https://user-images.githubusercontent.com/436483/141340744-dec7b960-5fa1-4458-89de-a885b6ddc22d.mov

After, with this PR

https://user-images.githubusercontent.com/436483/141340804-721e77b0-d5b0-4fe5-a500-5466e46af324.mov

### Considerations

This is yet another patch on the flysytem_s3 module, so this will make updates even more difficult to manage.

### Checklist

- [ ] Tested in Development

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -123,7 +123,8 @@
                 "Issue #2770263 - Amazon S3 CORS Upload like functionality.": "patches/flysystem_s3_cors_upload-2770263-60.patch",
                 "Issue #3219524 - CORS upload functionality doesn't work with Claro theme.": "patches/flysystem_s3-cors-upload-claro-theme-3219524-2.patch",
                 "Issue #3159928 - uriScheme() method missing after update to Drupal 9.": "patches/flysystem_replace-urischeme-method-3159928-with-2772847.patch",
-                "Issue #3248466 - Duplicate file entities when uploaded through CORS [DATA LOSS]": "patches/flysystem_3248466-fix-data-loss.patch"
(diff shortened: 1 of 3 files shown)


andrewrlee (approved):
LGTM

PR #300, opened 2021-11-23
Combine prison fields

leonk:
### Context

https://trello.com/c/CIICYklo/320-combine-prison-and-prison-category-fields
https://trello.com/c/o5gLfwTV/322-add-exclude-from-field
https://trello.com/c/zTewwsDx/323-tidy-up-all-content-so-that-if-it-all-associated-prisons-in-category-are-ticked-so-is-the-category
https://trello.com/c/XxUk8cje/287-update-all-content-with-weyland-cw-and-lindholme-and-adult-male-tagged-so-that-berwyn-is-excluded

(Yes 4 cards relate to this PR, unfortunately they must all be deployed together).

### Intent

This PR merges prison and prison category fields, by doing the following:
- Creates new terms for "Female", "Adult male", and "Youth male"  within the current prison taxonomy, and marks all the current prisons as child terms of these (according to how they are currently setup with the separate prison_categories taxonomy).
- Creates a new field on all content types/taxonomy terms that currently have the Prisons field.  This field will be used to copy across prison field values.  The reason for storing in a new field, is so the site can still work during the update.  It will also make it easier to rollback if there are any issues.
 Adds a "Exclude from prison" field, with added test coverage.  The exclude from field is minimised when editing/creating content, unless a value has already been set.
- Renamed prisoner_hub_entity_access to prisoner_hub_prison_access.  Updated to use the new prison field with hierarchy + exclude from field.
- Added a copy of the previous prisoner_hub_entity_access module.  This will stay enabled during the deployment, and uninstalled once content has been updated.  At this point we will switch to the new module.
- Updates all existing content + taxonomy terms to copy across values from previous prison and prison categories field, to the new merged prisons field.  In this process we also strip out unnecessary child prisons when the category has been selected.  We also add Berwyn as excluded when we are enabling the "Adult male" prison category but Berywn is not selected.

### Considerations
Yes this is a big PR :pensive:. All the changes of this PR need to be deployed together.  This PR updates all the content + taxonomy terms to use the new `field_prisons` and `field_exclude_from_prison`.  The only way to ensure that the correct content is shown is do this altogether (otherwise we would need to manage a state of using both the old and new fields).

However, it's mostly config changes (due to there are two fields being added, across lots of content types/taxonomies). To make it easier to review, you can deselect yml files from the diff.

Screenshot 2021-11-29 at 11 57 38

Also note that Github doesn't quite show how the prisoner_hub_entity_access was renamed to prisoner_hub_prison_access correctly.  Instead it looks like a lot of the files are new (which adds further noise to the diff).  Not really sure of a way around this, if you go into the history of a file, you can see the commit where it was renamed.

A follow up PR will be needed to update config with the newly installed module, and remove instances of the old field.

### Checklist

- [ ] Tested in Development


Makefile
```diff
@@ -22,6 +22,8 @@ push-preview:
 
 install-drupal:
 	vendor/bin/drush site-install prisoner_content_hub_profile --existing-config -y
+	vendor/bin/drush pm-uninstall prisoner_hub_entity_access
+	vendor/bin/drush pm-enable prisoner_hub_prison_access
```
leonk, line 26: Installing the new module manually (not via config) so that the new tests will pass on circleCI.
These two lines can be removed after this has been deployed.

composer.json
```diff
@@ -49,6 +49,7 @@
         "drupal/select2": "^1.13",
         "drupal/string_field_formatter": "^2.0",
         "drupal/taxonomy_machine_name": "1.x-dev",
+        "drupal/term_reference_tree": "^1.1",
```
leonk, line 52: This is new the module that provides the taxonomy tree functionality (i.e. select all child terms automatically when a parent term is selected).

config/sync/core.entity_form_display.node.allowlisted_website.default.yml
```diff
@@ -14,6 +14,7 @@ dependencies:
     - field.field.node.allowlisted_website.field_moj_top_level_categories
     - field.field.node.allowlisted_website.field_not_in_series
     - field.field.node.allowlisted_website.field_prison_categories
+    - field.field.node.allowlisted_website.field_prisons
```
leonk, line 17: The new prison field (previous one was called `field_moj_prisons`

docroot/modules/custom/prisoner_hub_prison_access/src/EventSubscriber/QueryAccessSubscriber.php
```diff
@@ -0,0 +1,128 @@
...
+   */
+  protected function getFieldBundles($entity_type_id, $field_name) {
+    $map = $this->entityFieldManager->getFieldMap();
+    if (isset($map[$entity_type_id][$field_name]['bundles'])) {
+      return $map[$entity_type_id][$field_name]['bundles'];
+    }
+    return [];
+  }
+
+}
```
leonk, line 128: This file is actually moved from the previous module. Which is retained in the git history.  But for some reason github doesn't show this as a modified file.

docroot/modules/custom/prisoner_hub_prison_access/src/SearchApiQueryAlter.php
```diff
@@ -0,0 +1,59 @@
...
+    $condition_group = new ConditionGroup('OR');
+    $condition_group->addCondition($this->prisonFieldName, $current_prison->id());
+
+    foreach ($current_prison->get('parent') as $parent) {
+      $condition_group->addCondition($this->prisonFieldName, $parent->target_id);
+    }
+
+    $query->addConditionGroup($condition_group);
+  }
+}
```
leonk, line 59: Same as above, it's actually a renamed file (not a new file).

composer.json
```diff
@@ -156,6 +157,10 @@
             },
             "drupal/entity_clone": {
                 "Issue #3036068: Make it possible to configure the published state of cloned entities.": "patches/entity_clone-3036068-20.patch"
+            },
+            "drupal/term_reference_tree": {
+                "Issue #2411667: \"Disabled\" state of the element is cleared.": "patches/term_reference_tree-2411667-disabled-state-of-element-is-cleared.patch",
+                "Issue #3251627: Add ability to enforce cascading selection..": "patches/term_reference_tree-3251627-enforce-selection.patch"
```
leonk, line 163: Patches provide the functionality of disabling prisons when a category has been selected.

andrewrlee (approved):
LGTM

andrewrlee (approved):
LGTM

PR #301, opened 2021-11-23
Adding copy of prisoner_hub_entity_access module

leonk:
### Context

This PR is based on https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/300

### Intent

This PR adds back in the `prisoner_hub_entity_access` module, which has been renamed to `prisoner_hub_prison_access` and modified to use the new prison field.  In https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/300

The version of the module in this PR is exactly the same as currently in main.  

The reason for adding this back in, is we want to have both versions of the module in the codebase.  We will keep this current version of the module enabled (which works with the old prison fields) so the site will still work normally during the update.  Once all the content has been bulk updated, the `prisoner_hub_entity_access`  module will be uninstalled and the `prisoner_hub_prison_access` installed.  At this point the access system will use the new prison field.

After this has been deployed, we can remove the `prisoner_hub_entity_access` module from the repo.

### Considerations

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #308, opened 2021-11-25
Bugfix/fix dev deployment

leonk:
### Context

No this is a small bug fix.

### Intent

The dev environment has become unstable, with pods recreating and picking up newly pushed (un-deployed tags).

This was due to the use of the `preview` tag.  Every branch pushed to github would build to this tag, so when pods recreated they would pick up the latest rather than the original they were on.

This PR updates the circleci build scripts to use a unique version for each build.  This should prevent the issue from happening.

The setup is copied from the frontend repo.

### Considerations

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #315, opened 2021-12-10
Block external jsonapi access + add new ingress

leonk:
### Context

> Does this issue have a Trello card?

https://trello.com/c/6vgxPaTU/492-only-allow-access-to-content-from-within-cloud-platform-network

### Intent

This PR adds a nginx configuration snippet that denies external access to /jsonapi routes.
After this update, only requests within cloud platform will be accepted through to any path starting with /jsonapi.

It also adds a new ingress + hostname, that is allowlisted to the MoJ network.  This does not include the nginx config to block `/jsonapi` routes, so can be used for local development + debugging.
The hostnames for each environment are:
- jsonapi-cms-prisoner-content-hub-development.apps.live.cloud-platform.service.justice.gov.uk
- jsonapi-cms-prisoner-content-hub-staging.apps.live-1.cloud-platform.service.justice.gov.uk
- jsonapi-cms-prisoner-content-hub-production.apps.live-1.cloud-platform.service.justice.gov.uk

### Checklist

- [ ] This PR contains **only** changes related to the above card
- [ ] Tests have been added/updated to cover the change
- [ ] Documentation has been updated where appropriate
- [ ] Tested in Development

PR #317, opened 2021-12-13
Simplify the trusted host setting

leonk:
### Context

This PR is based on https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/315

### Intent

This PR removes helper functions `prisoner-content-hub-backend.trustedHosts` and prisoner-content-hub-backend.trustedHostsJsonApi` which were difficult to read and created unnecessary configuration.

In order to remove this helper functions, we remove support for multiple hosts.  This allows us to pass through the single value directly into the env variable.

### Considerations

Removing support for multiple hosts should have no effect, as we are not using them anywhere.
Except for on production, where multiple hosts were defined: 
- manage.content-hub.prisoner.service.justice.gov.uk 
- cms-prisoner-content-hub-production.apps.live-1.cloud-platform.service.justice.gov.uk

Only the first is being used, so the second one was removed.  

### Checklist

- [ ] Tested in Development


leonk:
Closing this as it's now out of date with main.
Although it's still something we want to do.

PR #323, opened 2021-12-16
Allow homepage tiles to link to series

leonk:
### Context

https://trello.com/c/ouqhJ3R0/532-allow-homepage-tiles-to-link-to-series

### Intent

This PR adds two new fields to the homepage (featured_articles) content type.  These are exactly the same as the existing fields, but are dynamic entity reference fields (as opposed to standard entity reference fields).  This allows them to link to other entity types.  In this PR, we have enabled taxonomy terms and series.

The values from the previous two fields are copied across, and the previous two fields are then hidden from the form display.  So any content updates from this point will be to the new fields.

The two new fields are called:
- field_featured_tile_large (previously field_moj_featured_tile_large)
- field_featured_tile_small (previously field_moj_featured_tile_small)

### Considerations

This needs to be deployed _just before_ a change to the FE, that reads the new field names.  

After this PR has been deployed to prod, a second change should be made to remove the old fields.

### Checklist

- [ ] Tested in Development

### DIFF
--- a/config/sync/core.entity_form_display.node.featured_articles.default.yml
+++ b/config/sync/core.entity_form_display.node.featured_articles.default.yml
@@ -4,11 +4,14 @@ status: true
 dependencies:
   config:
     - field.field.node.featured_articles.field_exclude_from_prison
+    - field.field.node.featured_articles.field_featured_tile_large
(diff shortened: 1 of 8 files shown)


jonhindle (approved):
LGTM

PR #327, opened 2022-01-07
Fixing JS errors reported in Sentry by aggregating JS and CSS.

leonk:
### Context

https://sentry.io/organizations/ministryofjustice/issues/2846024128/?project=5377063&query=is%3Aunresolved

### Intent

This PR fixes JS errors being reported in Sentry.

These errors appear to be when JS is trying to access an object that is loaded from a separate JS file.  The only scenario that I can see this happening is if a network issue caused a file to fail loading.  In this case, the browser would still call the `document.ready` event and Drupal's JS would start to run.

This PR enables JS and CSS aggregation.  Which combines files into one (or sometimes into 3 or 4 files, depending on whether a grouping has been specified).  This is core Drupal functionality that is enabled on most Drupal sites.  Normally for performance reasons, but in our case this should help prevent the errors (as either all of the JS is loaded, or none at all).

### Considerations
Sometimes there can be issues enabling aggregation, if a css/js file does not finish in the correct way, which didn't have an effect before, but after combining with other js/css causes a syntax error.
This is quite rare, but something we should keep an eye out for.

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #328, opened 2022-01-11
Prevent page cache from being cleared during deployments.

leonk:
### Context

https://trello.com/c/LQ6xpxSV/581-fix-performance-issues-during-drupal-deployments

### Intent

This PR adds a new Drupal module that overrides the cache.page service, to prevent the page cache from being cleared when `drush cache:rebuild` is called.

This will mean the page cache is retained during deployments, fixing performance issues that effect the site immediately after a deployment (where Drupal has to rebuild the page cache for every request).

There is an alternative method for implicitly clearing the page cache.
```
drush cache:force-clear-page
```

Details of this will be added to https://dsdmoj.atlassian.net/wiki/spaces/HUB/pages/3757342835/Caching+in+Drupal

### Considerations

When Drupal clears out the cache, it also rebuilds the router, picking up any new or modified routes.
After this is done, it invalidates the entire page cache (it does not check to see whether there are actually any changes).  The module in this PR prevents the page cache from being invalidated.
This _should_ be fine, as each item in the page cache is stored with its url, so if a url on a route has changed, it will be considered a new cache entry (same goes for new routes, these will not exist as cache items).  
However, if there is a change within the router that does not effect the urls itself (e.g. an update to a parameter type on an existing route), the existing cache may prevent these changes from appearing immediately. For those situations, `drush cache:force-clear-page` will need to be called.

### Checklist

- [ ] Tested in Development

### DIFF
--- a/config/sync/core.extension.yml
+++ b/config/sync/core.extension.yml
@@ -51,6 +51,7 @@ module:
   node: 0
   options: 0
   page_cache: 0
+  page_cache_persist: 0
   path: 0
   path_alias: 0
   prisoner_hub_cms_help: 0
--- a/docroot/modules/custom/page_cache_persist/README.md
+++ b/docroot/modules/custom/page_cache_persist/README.md
@@ -1,0 +1,26 @@
+# Page Cache Persist
+
+### What
+This module makes Drupal's internal page cache persist when Drupal tries to clear all caches.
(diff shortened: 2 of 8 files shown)


andrewrlee (approved):
LGTM

PR #329, opened 2022-01-13
Fix redis cache clear on deployments

leonk:
### Context

https://trello.com/c/LQ6xpxSV/581-fix-performance-issues-during-drupal-deployments

### Intent

This an issues relating to the previous PR https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/328 where it was still re-generating the cache after each deployment.

It was discovered that Drupal's `hash_salt` was re-generated on each `helm upgrade`, and the value of this was being used for prefixing Redis cache keys.  This meant that everything in Redis would immediately become stale after each deployment.  

Here is an example of a Redis key, taken from the dev environment, this is for the page `https://jsonapi-cms-prisoner-content-hub-development.apps.live.cloud-platform.service.justice.gov.uk/jsonapi/node/page/decf2e6a-30b3-4df7-83cc-0c0fc76e9466?resourceVersion=id%3A3602`
```
"drupal.redis.9.3.0..c84ca7b5c170fcddcfd7f36ed6cdbee3d0c5ea5701f6879b3e8bde9471533076:page:https://jsonapi-cms-prisoner-content-hub-development.apps.live.cloud-platform.service.justice.gov.uk/jsonapi/node/page/decf2e6a-30b3-4df7-83cc-0c0fc76e9466?resourceVersion=id%3A3602:"
```
This is the code that generates the prefix (from docroot/core/lib/Drupal/Core/Site/Settings.php):
```
return 'drupal.' . $identifier . '.' . \Drupal::VERSION . '.' . static::get('deployment_identifier') . '.' . hash_hmac('sha256', $identifier, static::get('hash_salt') . '.' . $root . '/' . $site_path);
```

Drupal's `hash_salt` is used for a wide variety of tasks (the most common one being to generate one-time login links for forgotten passwords, which we aren't using, but there are plenty of other places where it's used, such as Redis).  The `hash_salt` should not be continuously updated, it's designed to be set once and left as is.

To fix this, this PR removes the secrets.yaml resource, which prevents the hash_salt being updated.
The existing secret has then been updated to use the `"helm.sh/resource-policy"=keep` policy (meaning that Helm will not remove it).
```
k -n prisoner-content-hub-development annotate --overwrite secrets prisoner-content-hub-backend "helm.sh/resource-policy"=keep
```

In addition to this, to remove the `drupal.redis.9.3.0` part (which will get changed everytime we update Drupal), a prefix has been set in settings.php
```
$settings['cache_prefix'] = 'prisoner_content_hub_backend';
```

After this change, the Redis key from above will look like this:
```
"prisoner_content_hub_backend:page:https://jsonapi-cms-prisoner-content-hub-development.apps.live.cloud-platform.service.justice.gov.uk/jsonapi/node/page/decf2e6a-30b3-4df7-83cc-0c0fc76e9466?resourceVersion=id%3A3602:"
```

### Considerations

The `k annotate` command from above needs to be run on staging and production environments before this is deployed.

We should also manually flush the redis cache on staging and prod (using redis-cli) immediately after the release, as there are many stale items in there due to the prefix issue.

### Checklist

- [ ] Tested in Development


andrewrlee (dismissed):
LGTM

andrewrlee (approved):
LGTM

PR #332, opened 2022-01-14
Updating patch to further prevent missing files in Drupal.

leonk:
### Context

Another one related to https://trello.com/c/s9Bvdh6L/728-fix-the-mysterious-missing-file-issue

### Intent

I spotted this issue just before Christmas, but didn't have time to create a PR for this.
The way the previous patch was checking for existing files wasn't working correctly 100% of the time.

The main part that was incorrect was:
```
->condition('uri', $basename, 'STARTS_WITH')
```
The `$basename` is actually the filename basename, e.g if a file is called `somerandomimage.jpg` the basename would be `somerandomimage`.
This wouldn't work in the check for existing files.  It needs to be 
```
->condition('uri', $directory, 'STARTS_WITH')
```
But there was also an issue with the way the files were originally being saved in Drupal.

### Considerations

This PR is showing a diff of a patch, which makes it quite difficult to read.
Managing all of the patch files for the flysystem_s3 module has become a real nightmare.  We now have 6 patches, and some must be generated on-top of others.
I've opened a request to become co-maintainer of the flysystem_s3 module, so that we can get some of the patches merged in.  https://www.drupal.org/project/flysystem_s3/issues/3258564 Should hopefully be a lot easier after this point.

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #334, opened 2022-01-14
Fixing "content has been modified by another user" error when re-saving content

leonk:
### Context

https://trello.com/c/e6MQjFUu/587-fix-this-content-has-been-modified-by-another-user-issue

### Intent

Previously, Drupal would redirect the user to the "View" page after saving content.  Since we disabled this page, the user is shown the same edit form they were previously on.
When content is saved, Drupal checks to see if a more recent change was made to the same content.  If so, it assumes that there is a more recent modification, and the current one is stale.  Showing the user the "The content has been modified by another user" error.
The date it checks from is when the edit form was originally loaded, in our case this was when the form was first loaded, i.e. the value wasn't being updated after the edit page is refreshed (when the content is saved). 
This could be considered a bug in Drupal, but because the scenario of the same content form being reloaded never occurs (as it would always redirect to another page), it's something we have run into with our own modifications.

The fix is to update the changed value on the form to the current time.

### Considerations

### Checklist

- [ ] Tested in Development

### DIFF
--- a/docroot/modules/custom/prisoner_hub_edit_only/prisoner_hub_edit_only.module
+++ b/docroot/modules/custom/prisoner_hub_edit_only/prisoner_hub_edit_only.module
@@ -42,11 +42,19 @@ function prisoner_hub_edit_only_form_alter(&$form, FormStateInterface $form_stat
 }
 
 /**
- * Submit handler that removes the redirect on node edit forms.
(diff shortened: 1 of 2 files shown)


andrewrlee (approved):
LGTM

PR #335, opened 2022-01-19
Allow lcms to use exclude from

leonk:
### Context

https://trello.com/c/RpvHoCwQ/574-allow-any-user-from-a-prison-to-add-own-prison-to-exclude-from-field-for-content
https://trello.com/c/mtYpgoDF/575-allow-any-user-from-a-prison-to-edit-content-published-by-anyone-from-that-prison

### Intent

This PR does the following:
- Adds a `field_prison_owner` to all content types (excluding external_link and help_page).  The bulk of the config yml updates are related to this
- Updates all existing content, looking at the author of each item, if that user is associated with a prison then it will assign the prison to the content (prisons have already been associated with users on production).
- Enables a new custom module `prisoner_hub_prison_access_cms` that sets form disabled states on all form fields when a user is not associated with the prison the content is owned by. (I.e. they can't make any changes).  Except for `field_exclude_from_prison` where they are allowed to add their own prison (but no others).
- Updates permissions to allow LCM's to edit any content on the site (excluding external_link and help_page).  This allows them to access the edit screen.
- Adds a permission to Studio admin role _"bypass prison ownership edit access"_, this means Studio admins are able to continue to make changes to all fields on all content types.

### Considerations

Currently the user -> prison mapping only exists on prod.  We should copy this down to staging/dev and test it there.

Also, a lot of this PR is config changes (due to having a add a new field to several content types).  To make it easier to read, exclude yml files from the diff. 

### Checklist

- [ ] Tested in Development


config/sync/field.field.user.user.field_user_prisons.yml
```diff
@@ -12,7 +12,7 @@ field_name: field_user_prisons
 entity_type: user
 bundle: user
 label: Prisons
-description: ''
+description: "The prison(s) the user is associated with.</br>\r\nThis will determine what content a user can make changes to.</br>"
```
andrewrlee, line 15: This will determine what content the user can change?

andrewrlee (dismissed):
lgtm

config/sync/field.field.user.user.field_user_prisons.yml
leonk: updated this

andrewrlee (approved):
LGTM

PR #337, opened 2022-01-20
Setting default value for release date field to current date.

leonk:
### Context

https://trello.com/c/18lNOxf4/603-fix-issue-with-same-content-re-appearing-on-pagination

### Intent

Note this won't actually fix the issue in the card.  For that a frontend change is required to the JSON:API sorting.

This PR sets the default value of the release date field to be the current date.  When testing the above issue, I found that there were lots of videos _without_ a release date.  Whilst this should not cause the issue, we should still set the default value to be the current date, so any new content will always have a release date.

### Considerations

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #339, opened 2022-02-07
Fix content suggestions not returning consistent results.

leonk:
### Context

https://trello.com/c/nbxMo7d4/635-you-might-like-component-is-bugged

### Intent

The you might like component was returning inconsistent results, with sometimes less than 4 items being returned inside `data`, and items appearing as `omitted`.

The issue was caused by the fact that unpublished items were not being excluded from the query.  This meant they got removed from the JSON:API response _after_ the query had run.  Resulting in a reduced total number of items being returned.

This PR adds the publishing status as a condition to the query, so that the 4 items returned by the query will always be accessible and not get excluded.

### Considerations

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #340, opened 2022-02-10
Bugfix/image upload button not working

leonk:
### Context

https://trello.com/c/48w0up7I/661-fix-image-upload-button-not-working-in-text-editor

### Intent

This PR fixes an issue where the image upload button in the text editor wouldn't initially work.

The issue was due to the fact that clicking the upload button triggers an ajax request, and this loads in an additional Javascript file.  For _some reason_ when this file is loaded from S3, it appears to be loaded in _after_ Drupal's calls `Drupal.attachBehaviors()` (which is like an event to say that something new has been loaded in).  Because of this, nothing happens.  When the additional Javascript file is loaded from the same hostname as the site is currently on, it's loaded in correctly (i.e. before `Drupal.attachBehaviors()`).

I couldn't work out why this is, and didn't see anyone else reporting the same issue.  

Rather than directly fix the issue, I decided to load the JS/CSS files directly via Drupal.  Using a new flysystem scheme, that streams the files from S3.  This has the additional benefit of not requiring S3 signatures, which also means the CSS/JS files can be cached in the users browser.

I had to also fix an additional issue where cache headers were not being set correctly, as part of this PR.

### Considerations

### Checklist

- [ ] Tested in Development

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -80,6 +80,9 @@ RUN pecl install uploadprogress \
 RUN pecl install redis \
   && docker-php-ext-enable redis
 
+# Enable apache modules that are used in Drupal's htaccess.
+RUN a2enmod expires headers
+
 RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" \
   && php composer-setup.php --install-dir=/bin --filename=composer --version=2.1.8 \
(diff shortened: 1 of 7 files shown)


andrewrlee (approved):
LGTM

PR #344, opened 2022-02-15
Fix broken images in text editor.

leonk:
### Context

https://trello.com/c/GQDgnT7h/658-fix-images-not-appearing-in-text-editor

### Intent

This PR adds a new custom module to fix the broken issue of broken images inside the text editor.

The issue was caused by the use of s3 presigned urls.  These do not get updated on the input text (only on output), so they would expire after 1 day and appear broken when editing content.

The fix was to apply the same processor that was already being applied on the output, which updated the urls (refreshing the signatures) on the input text.

I also tested this out with the new version of the text editor (ckeditor 5) and it still works there

### Considerations

### Checklist

- [ ] Tested in Development


jonhindle (approved):
lgtm

PR #345, opened 2022-02-16
Feature/restrict empty categories

leonk:
### Context

https://trello.com/c/COr4tjCj/659-dynamic-check-to-hide-taxonomy-levels-that-have-no-content-underneath-them

### Intent

This PR adds support for restricting access to categories when there is no available content.
The check is done in the same way as the existing check for series content.  But the check for _content within a category_ is slightly different:
Content can either be:
- Assigned to a series that's assigned to the category
- Assigned directly to the category.

### Considerations

An issue we may run into is that the check for series/category having available content is done _after_ the JSON:API count.  This means we may run into incomplete sets of content.  E.g. if you request 40 items, you may only get back 35.
Whilst this hasn't so far been an issue for series, once we move to sub-categories it may become one. 

I did have a quick look into how we would do that when doing these changes, and it does not look straightforward. 

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #349, opened 2022-02-18
Allow users to add/remove their own prison from the prison field.

leonk:
### Context

https://trello.com/c/JzCzE9Vz/688-allow-dcms-to-add-their-own-prison-to-any-content

### Intent

This PR extends the existing functionality, of allowing local content managers to add their own assigned prison to `field_exclude_from_prison`, applying the exact same rules to the `field_prisons`.

### Considerations
The only real difference between the two fields, is that in `field_prison` you can select prison categories (i.e parent terms), whereas with `field_exclude_from_prison` these can never be selected.
If a prison category is selected on content, all the prisons within this category will be checked and disabled.  If a user has access to one of those prisons, it will still be disabled.

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #362, opened 2022-03-04
Remove and clear featured on category field

leonk:
### Context

https://trello.com/c/pm60I56L/711-implementation-of-sub-categories-design

### Intent

This card updates how content is featured on categories.

- The `field_feature_on_category` is removed, as we no longer want the ability to feature items from the content edit page (instead the only way to feature items will be by editing the category directly).
- The additional functionality around allowing content to be featured on a category has been removed/uninstalled.  This includes the "Corresponding Entity References" (cer) module, which was used to synchronise the values of `field_feature_on_category` and `field_featured_tiles`.
- The `field_featured_tiles` is still retained, as this the actual true store of featured items.  But the values of this field is cleared.

### Considerations

There is quite significant (but intentional) data loss that this PR will bring once deployed.  A db backup should be taken just before and stored somewhere securely. 

Once this PR is deployed, there will be nothing featured on any category.  Again this is intentional, as we want to limit the amount of featured items, only use this functionality when necessary, and build the list back up from scratch.

If we want to restore the functionality removed by this module, it _may_ be possible to do a simple revert.  It's possible that the config yml files will have conflicts with the latest set of config.  In that case, it may be worth doing an "Accept-theirs" of the config (so that the nothing from what is in the main branch is wiped) and then manually creating the `field_feature_on_category` again.  The two modules `cer` and `prisoner_hub_featured_content` need to be re-installed, and the cer.corresponding_reference.category_featured_content.yml re-imported (although this should import fine as it won't have any conflicts).

### Checklist

- [ ] Tested in Development


jonhindle (approved):
lgtm

PR #365, opened 2022-03-07
Make cronjob more performant

leonk:
### Context

### Intent

During a recent issue with performance directly after a prod deployment, https://dsdmoj.atlassian.net/wiki/spaces/HUB/pages/3903488780/2022-03-07+Drupal+unresponsive
One of the changes we made was to disable the "search-indexing" cronjob.  As this runs every 5 minutes and uses up resources.

This PR updates the cronjob in two ways:
- Rename the job to just "drupal-cron" rather than "drupal-search-indexing". As actually, Drupal is configured to "track changes immediately", so it's not reliant on cron for indexing.  However, cron is still required to be run on Drupal for other tasks, so a generic name is more suitable.
- Run the job once every 2 hours instead of 5 mins.  (We could probably decrease this futher if we wanted).
- Use drush to run cron via the cli, rather than a curl request over http.

### Considerations

### Checklist

- [ ] Tested in Development

### DIFF
--- a/helm_deploy/prisoner-content-hub-backend/templates/cronjob.yaml
+++ b/helm_deploy/prisoner-content-hub-backend/templates/cronjob.yaml
@@ -1,7 +1,7 @@
 apiVersion: batch/v1beta1
 kind: CronJob
 metadata:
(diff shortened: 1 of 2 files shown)


andrewrlee (approved):
LGTM

PR #369, opened 2022-03-09
Redirect to content overview page after login.

leonk:
### Context

### Intent

Currently when you login to Drupal you are shown your account page, which is not particularly useful to the user (it just displays how long the use has had an account for, and nothing else).

This PR will redirect them to the content overview page, once they have logged in.

### Considerations

### Checklist

- [ ] Tested in Development


leonk:
Don't think we need this, can just set this homepage to be the /admin/content page and this will have the same effect.

PR #374, opened 2022-03-16
Primary nav

leonk:
### Context

https://trello.com/c/KlX1FivQ/753-implement-primary-nav-bar

### Intent

This PR adds support for defining the primary navigation for each prison in Drupal. 
This is done using the Drupal menu system.

The menu can be retrieved via JSON:API, using 
`/jsonapi/prison/{prison}/primary_navigation`

A default menu is setup (as part of this PR), this will be used as the fallback, if a prison hasn't selected their own menu.
To create a prison-specific menu, firstly the menu will need to be created in Drupal:
Screenshot 2022-03-21 at 10 26 28
Screenshot 2022-03-21 at 10 27 54
Next edit the prison and select the menu that was just created as the primary navigation.
Screenshot 2022-03-21 at 10 29 35

### Considerations

Because a menu in Drupal is considered configuration, which gets exported as yml.  To avoid unnecessary config additional config yml files, and indadvertedly removing menu's from Drupal on deployments. The config_ignore module is used to ignore any newly created menus.

### Checklist

- [ ] Tested in Development


jonhindle (approved):
lgtm

PR #379, opened 2022-03-22
Rename tags to topics

leonk:
### Context

https://trello.com/c/V1TQtL3I/771-rename-secondary-tags-to-topics-in-drupal

### Intent

This PR:
- Creates a new "topics" taxonomy
- Converts all secondary tags to topics (meaning they retain the same ID's as before
- Adds a new `field_topics` to all content types that had the secondary tag field.
- Copies over all the values from `field_moj_secondary_tags` to `field_topics`
- Updates all references to secondary tag in any custom code.

### Considerations

Because this PR converts the tags (rather than copying) it will need to be deployed at the same time as an update to the frontend (to use topics instead of tags).  As there will be no secondary tags left on the site after this PR has been deployed.

A subsequent "cleanup" PR will be required to remove the old field and taxonomy type.

### Checklist

- [ ] Tested in Development


jonhindle (approved):
lgtm

PR #382, opened 2022-03-24
Making the alt tag field required on thumbnail images.

leonk:
### Context

### Intent

This PR enables the alt tag field for thumbnail images on the audio and video content type.
It also makes this field required.

Previously this field was not enabled, meaning we are not capturing and outputting alt tag attributes for images on video and audio items.

### Considerations

Once the alt tag is required, you won't be able to update existing content without adding it.  Even if you were trying to update something unrelated to the image.

### Checklist

- [ ] Tested in Development


davidthomas22 (approved):
LGTM

andrewrlee (approved):
LGTM

PR #386, opened 2022-03-31
Fixing 500 error when a referenced entity is deleted.

leonk:
### Context

No this is a small change.

### Intent

Yesterday an error was discovered on prod, a series had been deleted that was being referenced from the Berwyn homepage.
When this happens, the referencing content should still work, but just not load in the deleted item.  However, the page crashed. There was a 500 error coming from the Drupal JSON:API response.

The issue was caused by the fact we are using a patch on Drupal core to allow it to support the `dynamic_entity_reference` module (which is used on the homepage so you can reference both content and series/categories).  The patch did not work with deleted entities.

I've updated the patch to fix the issue.
See https://www.drupal.org/project/drupal/issues/3057545#comment-14464137

### Considerations

Lots of the diff on this PR are because the patch had to be updated to work with the latest version of Drupal core.  So they aren't real changes.
The only _real_ change is the following part
```
+            if ($field_item->entity instanceof EntityInterface) {
+              // Support entity reference fields, which don't have the referenced
+              // target type stored in settings.
+              $references[$field_item->entity->getEntityTypeId()][] = $field_item->get($field_item::mainPropertyName())->getValue();
+            }
+          }
```

### Checklist

- [ ] Tested in Development

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -136,7 +136,7 @@
             "drupal/core": {
                 "Issue #1149078 - States API doesn't work with multiple select fields": "patches/drupal_states-multiselect-1149078-109.patch",
                 "Issue #3228329 - Fix cache dependencies not set for files uploaded through text editor.": "patches/drupal-3228329-files-uploaded-through-text-editor-cache-dependency-2.patch",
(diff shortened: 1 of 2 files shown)


andrewrlee (approved):
LGTM

PR #387, opened 2022-03-31
Setting the content overview page as the homepage.

leonk:
### Context

No this is a small change.

### Intent

This sets the homepage in Drupal as the /admin/content page.  This means when a user logs in, they are immediately taken to this page (rather than the /user page).

### Considerations

The previous homepage was /user/login, which presented a logged out user with a login form.  
However as the /admin/content page is not accessible when logged out, the user will still see the login form (so it will appear the same as it did before).  There is a small difference:

Before:
Screenshot 2022-03-31 at 12 45 58
After:
Screenshot 2022-03-31 at 12 46 01

### Checklist

- [ ] Tested in Development


andrewrlee (approved):
LGTM

PR #390, opened 2022-04-28
Adding jsonapi resource for sub terms.

leonk:
### Context

> Does this issue have a Trello card?

https://trello.com/c/Ar0dte1r/820-change-sorting-of-subcats-series-to-show-relevant-order-in-specific-prisons

### Intent

Provides a new custom jsonapi resource for sub-categories and series:
`/jsonapi/prison/{prison name}}/taxonomy_term/moj_categories/{uuid}/sub_terms`
This is ordered by the most recently updated content (in the current prison).

### Considerations

The current method for retrieving sub-categories and series will continue to work.  A subsequent PR is needed to clean up the old method for sorting by content updated date (as there were performance issues related to this).  

### Checklist

- [ ] This PR contains **only** changes related to the above card
- [ ] Tests have been added/updated to cover the change
- [ ] Documentation has been updated where appropriate
- [ ] Tested in Development

PR #391, opened 2022-04-28
Remove the content_updated field which will soon no longer be used.

leonk:
### Context

https://trello.com/c/Ar0dte1r/820-change-sorting-of-subcats-series-to-show-relevant-order-in-specific-prisons

### Intent

After https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/390 is deployed and the fe is updated to use the new resource.  We can run this cleanup PR to remove the old functionality.

We should do this asap as it will improve performance (due to the fact that category taxonomy terms were being updated every time a node was updated, causing unnecessary cache clearing).

### Considerations

### Checklist

- [ ] Tested in Development


davidthomas22 (approved):
LGTM

PR #396, opened 2022-05-06
Removing permissions to use bulk update tool

leonk:
### Context

No this is a small change.

### Intent

Currently, DCM's (local content managers) can only edit content owned by their own prison.
However, the bulk update tool circumnavigates these restrictions, so they could use this to publish content that was owned by another prison.

The reason for this is that the restrictions on editing content are only placed on the content edit form, and are not global permissions.

### Considerations

### Checklist

- [ ] Tested in Development


davidthomas22 (approved):
LGTM

PR #393, opened 2022-05-04
Adding publication_date module

leonk:
### Context

> Does this issue have a Trello card?

This functionality will be used in a few upcoming Trello cards, the first one being 
https://trello.com/c/KRCuHwwR/924-create-a-page-that-shows-all-the-most-recent-content-added-to-the-hub

### Intent

Installs the [publication_date](https://www.drupal.org/project/publication_date) Drupal module, with patch from https://www.drupal.org/project/publication_date/issues/2983348#comment-13849551 to add dates for existing content.

The publication date becomes a new field, available in JSON:API as `published_at`. 
Screenshot 2022-05-04 at 11 06 00
This can be used for sorting.

The publication date field itself looks like this on the content edit form.

Screenshot 2022-05-04 at 10 42 27

### Considerations

> Is there any additional information that would help when reviewing this PR?

> Are there any steps required when merging/deploying this PR?

### Checklist

- [ ] This PR contains **only** changes related to the above card
- [ ] Tests have been added/updated to cover the change
- [ ] Documentation has been updated where appropriate
- [ ] Tested in Development

PR #397, opened 2022-05-10
Fixing failed publication date module install.

leonk:
### Context

Fixes issue with deployment of https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/393

### Intent

The above PR installed the publication_date module, this failed to install properly on prod.  The error was 
```
Error Output:
  ================

  In DiscoveryTrait.php line 53:


    The "field_item:published_at" plugin does not exist
```
I'm not 100% sure why this happened (and only occurred on prod).  It looks as though the post-install hook ran too early, and the code wasn't there (as the plugin mentioned above should have been available).  Possibly there is a cache of the plugins that wasn't cleared in time.

Either way, the issue now is that the module is technically installed, but it didn't create the necessary column in the database, and didn't run the update to populate the data for existing content.

This PR adds an update hook to ensure the necessary column is created, the updates from the publication_date module will then run without any issues.

### Considerations

### Checklist

- [ ] Tested in Development

### DIFF
--- a/docroot/profiles/prisoner_content_hub_profile/prisoner_content_hub_profile.install
+++ b/docroot/profiles/prisoner_content_hub_profile/prisoner_content_hub_profile.install
@@ -312,3 +312,11 @@ function prisoner_content_hub_profile_update_8020(&$sandbox) {
   return 'Processed terms: ' . $sandbox['progress'];
 
 }
+
+/**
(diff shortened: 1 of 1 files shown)


jonhindle (approved):
lgtm

PR #398, opened 2022-05-11
Feature/recently added resource

leonk:
### Context

https://trello.com/c/LjGu2N1C/920-allow-users-to-see-all-recently-added-content-on-the-homepage

### Intent

Adds a new custom jsonapi resource:
`/jsonapi/recently-added`.

This displays the most recently published content.
If content is in a series, it displays the series instead (and removes duplicates of the same series).
If content is not in a series, it displays the content.

Note this resource doesn't fully support pagination.  However, it does support a limit, so you should use it like:
`/jsonapi/prison/wayland/recently-added?page[limit]=4`

### Considerations

The `field_featured_image` on taxonomy terms has been renamed to `field_moj_thumbnail_image`.
This will allow us to add this as an include to the resource, i.e.
`/jsonapi/prison/wayland/recently-added?page[limit]=4&include=field_moj_thumbnail_image`

Otherwise if the field isn't present on both nodes and taxonomy terms you get an error.

This means that all references to `field_featured_image in the f/e code will need to be updated.  When this PR is deployed the old field will still exist (and should be removed in a separate PR).  But any new updates to an image on a taxonomy term will go to the new field.

### Checklist

- [ ] Tested in Development

### DIFF
--- a/config/sync/core.entity_form_display.taxonomy_term.moj_categories.default.yml
+++ b/config/sync/core.entity_form_display.taxonomy_term.moj_categories.default.yml
@@ -6,6 +6,7 @@ dependencies:
     - field.field.taxonomy_term.moj_categories.field_exclude_from_prison
     - field.field.taxonomy_term.moj_categories.field_featured_image
     - field.field.taxonomy_term.moj_categories.field_featured_tiles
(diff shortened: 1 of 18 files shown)


andrewrlee (approved):
LGTM

BarryGeeMOJ (approved):
LGTM

PR #410, opened 2022-06-06
Feature/readonly mode

leonk:
### Context

https://trello.com/c/wqt78xna/1008-allow-dcms-and-studio-admins-to-continue-using-drupal-during-deployments

### Intent

This PR adds a new module `readonlymode`, which, when enabled, ensures no content can be updated.
By using this new mode, we can allow DCMs and studio admins access to the site when in maintenance mode (i.e. during deployments).  As the reason Drupal is placed in maintenance mode is that updates to the database could be taking place, so any content saved during this time can cause issues.

The deployment is updated to explicitly enable both maintenance mode and readonly mode before the deployment starts, and disabled once it has finished (previously Drupal would enable maintenance mode automatically, but we want to ensure it's enabled at the same time as readonly mode, so we do it ourselves).

### Considerations

Version 2.x of the readonlymode module is being used, as version 1 had a method of hiding the entire content edit form.  Whereas with version 2 you can still see and edit the form, but get a validation error when saving.

Below are the messages displayed when the site is in a deployment:
Screenshot 2022-06-06 at 13 22 13

When attempting to save content:
Screenshot 2022-06-06 at 16 24 39

### Checklist

- [ ] Tested in Development

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,5 @@
-build-deps:
-	composer clear-cache
-	composer install --no-dev --no-ansi --no-scripts --prefer-dist --ignore-platform-reqs --no-interaction --no-autoloader
-
 build:
 	docker build -t prisoner-content-hub-backend .
-
-clean:
-	rm -rf modules/contrib
 
 push:
 	@docker login -u $(DOCKER_USERNAME) -p $(DOCKER_PASSWORD)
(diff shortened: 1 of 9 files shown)


jonhindle (approved):
lgtm

PR #415, opened 2022-06-21
Set taxonomy view pages

leonk:
### Context

https://trello.com/c/HK5LEtQ4/1029-help-dcms-understand-how-the-new-homepage-updates-section-works

### Intent

As part of the card above, we are going to give DCM's access to the taxonomy overview page.  But the issue is they cannot edit taxonomy terms, so all they will be shown on the taxonomy overview page is a list of names and nothing to do with them (i.e. nothing they can click on).

This PR gives taxonomy a "view" page (which was previously removed, and the user was taken straight to the edit page).
The view page shows a list of content inside each taxonomy.  DCM's will be able to access this but not the edit page.  This will allow them to find content inside a taxonomy.

For Series and Topics, the "Change episode order" page is used - these pages show the order of content as it would be shown on the f/e, and allow the relevant fields that effect the order to be updated in bulk, so the order can be easily changed.

Series:
Screenshot 2022-06-21 at 11 57 23

Topics:
Screenshot 2022-06-21 at 11 57 41

Categories look like this:
Screenshot 2022-06-21 at 11 57 18
Note that the "Subcategory" filter shows a list of subcategories of the current category.  This allows the user to see what subcategories are available.

### Considerations

This really is a MVP and we will definitely need to re-visit some of these pages later.  Particularly the category page, as this will have the most content, and we are currently not showing any series content (or a list of series inside the category).

Also, as the "view" page of a Taxonomy has now been made available, this means it could be access when logged out.  Previously it would take you to the edit page, which anonymous users do not have access to, so Drupal would serve a 403.  But as anonymous users can view a Taxonomy (which they do through JSON:API), this means the new pages in Drupal also became available. 
To fix this, an extra permission has been added:
`view entity without prison context`
Without this permission, a user cannot view an entity without there being a prison context in the url.  And currently the only urls that have prison contexts are JSON:API urls.  This therefore means anonymous users can only access taxonomy through JSON:API.  Whereas DCM's will have this permission, so can access through Drupal (when they are logged in).

### Checklist

- [ ] Tested in Development


BarryGeeMOJ (commented):
LGTM

BarryGeeMOJ (approved):
LGTM

PR #427, opened 2022-07-15
Updating drupal core to 9.4.3, with dependencies and new config.

leonk:
### Context

### Intent

Update to latest stable version of Drupal core.  Part of the reason for doing this is to get the latest version of the new text editor (ckeditor5).  But also for general house keeping.

### Considerations
The additional changes to config and settings file were automatically generated by the update (so need to be pushed into git to avoid having changes appearing later on).

### Checklist

- [ ] Tested in Development


BarryGeeMOJ (dismissed):
lgtm

PR #433, opened 2022-08-16
Fix absolute urls being used in the text editor.

leonk:
### Context

https://trello.com/c/ZTW15U4f/1387-prevent-urls-that-wont-work-from-being-used-in-the-texteditor

### Intent

Installs a module called `pathologic` which adds a text filter to update urls.  The configuration has been setup to convert urls to be relative.  Therefore urls like `https://wayland.content-hub.prisoner.service.justice.gov.uk/tags/1282` will become `/tags/1282`.

Also enables a module called `linkit` which allows the user to search for the title of content/taxonomy instead of copy pasting the url.

Screenshot 2022-08-16 at 11 22 49
Screenshot 2022-08-16 at 11 20 10

### Considerations

### Checklist

- [ ] Tested in Development

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,7 @@
         "drupal/jsonapi_menu_items": "^1.2",
         "drupal/jsonapi_page_limit": "^1.0@beta",
         "drupal/jsonapi_search_api": "^1.0@RC",
+        "drupal/linkit": "^6.0",
         "drupal/maintenance_exempt": "1.x-dev",
         "drupal/memcache": "^2.0",
         "drupal/migrate_tools": "^5.0",
@@ -44,6 +45,7 @@
         "drupal/node_edit_protection": "^1.0",
         "drupal/optgroup_taxonomy_select": "^1.2",
         "drupal/pathauto": "^1.8",
+        "drupal/pathologic": "^1.0@alpha",
         "drupal/profile_switcher": "1.0-alpha5",
         "drupal/publication_date": "^2.0@beta",
         "drupal/raven": "^3.2",
@@ -185,6 +187,9 @@
             },
             "drupal/publication_date": {
                 "Issue #2983348: Published on empty for existing nodes.": "patches/publication_date-fill_published_at-2983348-60-for-composer.patch"
+            },
+            "drupal/pathologic": {
+                "Issue #758118: wild card for Additional paths.": "patches/pathologic-758118-wildcard-for-additional-paths-18.patch"
             }
         },
         "patches-ignore": {
--- a/composer.json
+++ b/composer.json
@@ -190,6 +190,9 @@
             },
             "drupal/pathologic": {
                 "Issue #758118: wild card for Additional paths.": "patches/pathologic-758118-wildcard-for-additional-paths-18.patch"
+            },
+            "drupal/linkit": {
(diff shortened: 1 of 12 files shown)


config/sync/filter.format.basic_html.yml
```diff
@@ -16,7 +18,7 @@ filters:
     status: true
     weight: -10
     settings:
-      allowed_html: '<a href hreflang> <em> <strong> <cite> <blockquote cite> <code> <ul type> <ol start type> <li> <dl> <dt> <dd> <h2 id> <h3 id> <h4 id> <h5 id> <h6 id> <p> <br> <span> <img src alt height width data-entity-type data-entity-uuid data-align data-caption> <panel>'
+      allowed_html: '<em> <strong> <cite> <blockquote cite> <code> <ul type> <ol start type> <li> <dl> <dt> <dd> <h2 id> <h3 id> <h4 id> <h5 id> <h6 id> <p> <br> <span> <img src alt height width data-entity-type data-entity-uuid data-align data-caption> <panel> <a href hreflang data-entity-substitution data-entity-type data-entity-uuid title>'
```
leonk, line 21: Not quite sure what caused this to change.  Must have been one of the modules that got enabled.  I went through the change and it doesn't look like anything has actually been added or removed, just a slight re-arrangement.

BarryGeeMOJ (dismissed):
lgtm

composer.json
```diff
@@ -185,6 +187,12 @@
             },
             "drupal/publication_date": {
                 "Issue #2983348: Published on empty for existing nodes.": "patches/publication_date-fill_published_at-2983348-60-for-composer.patch"
+            },
+            "drupal/pathologic": {
+                "Issue #758118: wild card for Additional paths.": "patches/pathologic-758118-wildcard-for-additional-paths-18.patch"
```
leonk, line 192: This patch is needed so that we don't have to specify each full prison hostname.  Instead we can use `https://*.content-hub.prisoner.service.justice.gov.uk`

The pathologic module needs to know which hostnames to convert (from absolute to relative).  It won't just convert everything.

composer.json
```diff
@@ -185,6 +187,12 @@
...
+            },
+            "drupal/linkit": {
+                "Issue #2877535: Link shown after the autocomplete selection is the bare node/xxx link, not the alias": "patches/linkit-2877535-link-shown-after-autocomplete-is-bare-41.patch"
```
leonk, line 195: This patch is for the link browser.  The default behaviour is to show the internal url to the user after they have selected a link, e.g. `/node/123`.  But our users won't know about these types of urls, and only know the `content/123` urls, so the patch updates the link browser to show them this instead.

config/sync/filter.format.full_html.yml
```diff
@@ -14,32 +16,74 @@ filters:
...
+    weight: -41
+    settings:
+      settings_source: global
+      local_settings:
+        protocol_style: full
+        local_paths: ''
+  filter_autop:
+    id: filter_autop
+    provider: filter
+    status: false
```
leonk, line 61: For some reason, it's exported the config with these new filters.  But as they are `status: false` it means they're not enabled.

BarryGeeMOJ (approved):
lgtm

PR #438, opened 2022-09-06
Adding new db backup cronjob.

leonk:
### Context

> Does this issue have a Trello card?

https://trello.com/c/JnaHKTyN/758-improve-drupal-database-backups

### Intent

Adds a nightly cronjob to backup the production database to S3.
The backups go to a new dedicated S3 bucket.  Already setup in https://github.com/ministryofjustice/cloud-platform-environments/pull/8543

### Considerations

> Is there any additional information that would help when reviewing this PR?

> Are there any steps required when merging/deploying this PR?

### Checklist

- [ ] This PR contains **only** changes related to the above card
- [ ] Tests have been added/updated to cover the change
- [ ] Documentation has been updated where appropriate
- [ ] Tested in Development

PR #439, opened 2022-09-07
Database refresh

leonk:
### Context

https://trello.com/c/BTPRqFGj/1386-automate-database-migration

### Intent
This PR adds automated database refreshing via k8ns cronjobs.  The setup is _a little bit_ similar to https://github.com/ministryofjustice/use-of-force/tree/main/database_refresh and https://github.com/ministryofjustice/hmpps-helm-charts/blob/main/charts/generic-service/templates/prison-postgres-database-restore.yaml which both do a similar thing on other services.  **Except** ours is a little more complicated as we also need to run the Drupal deployment cli command (`drush deploy`) _after_ the db has been imported.  Therefore a different approach is taken, whereby the cronjob is run on the dev and staging environments themselves (whereas on the other two projects the cronjob is run directly on prod).  This allows us to run the `drush deploy` command.

In order to get the database from prod to dev/staging, a separate db-backup cronjob is run (on production).  This backups the database to a dedicated S3 bucket (outside of the prod namespace) which dev/staging can then retrieve from.

Database backups are already added in a separate PR https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/438

### Considerations

Add a DB refresh script, which is run on cron (at 4.30am every day) and on the helm post install hook.

### Checklist

- [ ] Tested in Development


leonk:
Temporarily closing as the parent branch needs to be merged in first.

helm_deploy/prisoner-content-hub-backend/templates/db-refresh.yaml
```diff
@@ -0,0 +1,54 @@
...
+{{ .Files.Get "db-refresh.sh" | indent 4 }}
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ include "prisoner-content-hub-backend.fullname" . }}-db-refresh-cronjob
+  labels:
+    {{- include "prisoner-content-hub-backend.labels" . | nindent 4 }}
+spec:
+  schedule: "30 4 * * *"
```
leonk, line 20: Note this is scheduled to run 30 mins after the db backup cronjob. https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/438/files#diff-404543f25cde5fb7ecafaa0a26e02153e3356cdfe443b158a6d038a80e5e0093R20

helm_deploy/prisoner-content-hub-backend/templates/db-refresh.yaml
```diff
@@ -0,0 +1,54 @@
...
+          - name: db-refresh-cronjob
+            image: "ghcr.io/ministryofjustice/hmpps-mysql-tools:latest"
+            command:
+              - /bin/entrypoint.sh
+            volumeMounts:
+              - name: db-refresh-script
+                mountPath: /bin/entrypoint.sh
+                readOnly: true
+                subPath: entrypoint.sh
+  {{ include "db-backup.envs" . | nindent 12 }}
```
leonk, line 42: The db refresh container is run as an `initContainer`.  This means it runs first, before the deployment.

The one thing to be aware of this method, is that it can be quite hard to obtain logs if something fails within the initContainer.  If you do `kubectl logs <pod-name>` you will just see that it failed to startup and no further messages.  To get the actual messages you need to do: `kubectl logs <pod-name> -c <init-container-2>`. See https://kubernetes.io/docs/tasks/debug/debug-application/debug-init-containers/

helm_deploy/prisoner-content-hub-backend/templates/drupal-post-install-hook.yaml
```diff
@@ -18,6 +18,24 @@ spec:
...
+        - name: db-refresh
+          image: "ghcr.io/ministryofjustice/hmpps-mysql-tools:latest"
+          command:
+            - /bin/entrypoint.sh
+          volumeMounts:
+            - name: db-refresh-script
+              mountPath: /bin/entrypoint.sh
+              readOnly: true
+              subPath: entrypoint.sh
+{{ include "db-backup.envs" . | nindent 10 }}
```
leonk, line 32: We also run the db refresh on deployment, as part of the post-install-hook.  Note this only runs where   This ensures we are always testing where `.Values.dbRefresh.enabled` is set, which is just on dev and staging.
Running the db import at this point will slow down the deployment process, but ensures we are always testing our changes with the most up to date database.

BarryGeeMOJ (approved):
lgtm

PR #440, opened 2022-09-07
Improve healthcheck performance

leonk:
### Context

https://trello.com/c/3U2JomOi/1584-improve-performance-of-the-health-check

### Intent

Remove previous health check - which was provided by a module in Drupal.
Instead use a simple static file served directly from Apache.  This has no dependency on the database, and does not make a db connection.

### Considerations

### Checklist

- [ ] Tested in Development


Dockerfile
```diff
@@ -106,11 +106,8 @@ COPY composer.json composer.lock ./
 COPY patches/ patches/
 
 # Copy Project
-COPY docroot/modules/custom docroot/modules/custom
-COPY docroot/themes/custom docroot/themes/custom
-COPY docroot/profiles docroot/profiles
+COPY docroot/ docroot/
```
leonk, line 109: Needed to copy the new `health` directory over, and figured it was easier/better to copy the whole directory.  Everything that's in docroot should be made available, as Drupal expects this folder to be published.

PR #444, opened 2022-09-08
Decrease timeout to 30 seconds, to match the timeout on the frontend.

leonk:
### Context

No this is a small change.  Not 100% sure we want do this yet, up for discussion.

### Intent

The Apache timeout directive has a default of 60 seconds https://httpd.apache.org/docs/2.4/mod/core.html#timeout
For some reason ours is set to 5 minutes (not sure why it was increased).

This PR sets it to 30 seconds, to match the timeout from the frontend.

### Considerations

Are there any requests that take longer than 30 seconds?

There shouldn't be, at least not from the frontend.
There are some cli requests that do, but these do not run through Apache, so won't be subject to this timeout.

Possibly there are some long running processes in the CMS?  But anything that takes over 30 seconds should really be split into a batch.

### Checklist

- [ ] Tested in Development

### DIFF
--- a/apache/apache2.conf
+++ b/apache/apache2.conf
@@ -89,7 +89,7 @@ PidFile ${APACHE_PID_FILE}
 #
 # Timeout: The number of seconds before receives and sends time out.
 #
-Timeout 300
+Timeout 30
 
 #
 # KeepAlive: Whether or not to allow persistent connections (more than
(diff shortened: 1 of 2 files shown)


jonhindle (approved):
lgtm

BarryGeeMOJ (approved):
lgtm

PR #446, opened 2022-09-12
Re-exporting config that Drupal seems to want to always re-order.

leonk:
### Context

No this is a small change.

### Intent

For some reason Drupal wants to re-order this config, so it's re-importing it on every deployment.

This PR adds it in the correct format, so that it doesn't get picked up as a new config change.

### Considerations

### Checklist

- [ ] Tested in Development


BarryGeeMOJ (approved):
lgtm

PR #447, opened 2022-09-20
Improve performance by reducing unnecessary cache invalidations.

leonk:
### Context

https://trello.com/c/g4cjHCyT/1585-prevent-certain-jsonapi-requests-from-having-their-cache-regularly-cleared

### Intent

Fix performance issues when content is updated during peak hours.  The issue we've been seeing is that due to Drupal's default use of the generic `node_list` cache tag, this results in any content updates invalidating all series/topics pages, the content component on category pages, and the updates section on the homepage.  When done during high traffic periods, Drupal has to rebuild the cache for all of these pages at the same time.

This PR adds a custom module that adds cache tags based on jsonapi filter values.  So we only end up invalidating the requests we need to.

For more info on cache tags see https://dsdmoj.atlassian.net/wiki/spaces/HUB/pages/3757342835/Caching+in+Drupal (which also should be updated with details of this change).

### Considerations

- This module is designed to only alter the output of the cache tag in certain situations (see the limitations section of the module README.md).  So it won't apply everywhere, in particular the homepage updates component (as this uses a group filter).
- Only the `node_list` cache tag is removed, the individual node cache tags that are part of the response are still kept in, e.g. `node:17682`.  This is the correct functionality as we want any updates to these nodes to also invalidate the cache (it could be that they were changed to another category or series, so it wouldn't then trigger the cache tag invalidation from this module). 
- Tests still need to be written, will do that as part of the firebreak week.

### Checklist

- [ ] Tested in Development

### DIFF
--- a/config/sync/core.extension.yml
+++ b/config/sync/core.extension.yml
@@ -46,6 +46,7 @@ module:
   jquery_ui_touch_punch: 0
   jsonapi: 0
   jsonapi_cross_bundles: 0
+  jsonapi_filter_cachetags: 0
   jsonapi_image_styles: 0
   jsonapi_menu_items: 0
   jsonapi_page_limit: 0
--- a/config/sync/core.extension.yml
+++ b/config/sync/core.extension.yml
@@ -46,7 +46,7 @@ module:
   jquery_ui_touch_punch: 0
   jsonapi: 0
   jsonapi_cross_bundles: 0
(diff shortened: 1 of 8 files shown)


BarryGeeMOJ (dismissed):
lgtm

BarryGeeMOJ (approved):
lgtm

PR #454, opened 2022-10-04
Run docker build (composer etc) as non root

leonk:
### Context

https://trello.com/c/9XtYmHgl/1442-update-php-version-to-8-as-74-is-eol-in-november-2022

### Intent

As prep for the above card, update the Dockerfile to run composer as a different user from root.
As well as being good security practise, this should also speed up the builds - as previously we were having to run `chown` on the entire directory after running composer, which would take quite a while.

### Considerations

### Checklist

- [ ] Tested in Development


.circleci/config.yml
```diff
@@ -60,7 +60,7 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker:
-          version: 19.03.13
+          version: 20.10.17
```
leonk, line 63: Updating to latest version of docker in circleci

Dockerfile
```diff
@@ -129,27 +130,26 @@ COPY phpunit.xml phpunit.xml
...
-# Install dependencies (with dev)
+# Set to www-data user.
+USER 33
+
+RUN mkdir -p ~/phpunit/browser_output
+
+## Install dependencies (with dev)
 RUN composer install \
   --no-ansi \
-  --dev \
```
leonk, line 135: --dev is now the default so removing it here.

composer.json
```diff
@@ -78,7 +78,17 @@
...
+        "sort-packages": true,
+        "allow-plugins": {
+            "composer/installers": true,
+            "drupal/core-composer-scaffold": true,
+            "drupal/core-project-message": true,
+            "cweagans/composer-patches": true,
+            "drupal/console-extend-plugin": true,
+            "drupal/core-vendor-hardening": true,
+            "oomphinc/composer-installers-extender": true,
+            "dealerdirect/phpcodesniffer-composer-installer": true
```
leonk, line 90: `allow-plugins` is now required in the latest version of composer.

BarryGeeMOJ (approved):
lgtm

BarryGeeMOJ (approved):
lgtm

PR #463, opened 2022-10-06
Fix issue with topics not automatically getting a url.

leonk:
### Context

No this is a small change.

### Intent

- Hide the path field from the topics create/edit form (this is hidden on all other taxonomy and content types, as the path is always automatically generated).
- Update the path generation rule to include topics (it was previously targeting secondary tags, and wasn't updated when we switched over to topics).

### Considerations

### Checklist

- [ ] Tested in Development


BarryGeeMOJ (approved):
lgtm

PR #466, opened 2022-10-20
Remove the (now unused) secondary tag field.

leonk:
### Context

There was one somewhere, from a long time ago.

### Intent

When we switched from "Secondary tags" to "Topics".  We also renamed the field in Drupal. We did this by creating a new field, copying the data over, and then "hiding" the old field (this is the only safe and reliable way to rename fields).  

A PR should have been opened at the time to remove the old field, but this never happened.

### Considerations

Unfortunately, this PR shows just how much config Drupal can spit out for just removing a field.  There are also some unrelated additional bits of config updates that have made it in, this is due to the fact that re-saving configuration in Drupal can bring in new parts of config from other modules, that wouldn't have been enabled the last time it was saved. (It's really down to bad practise from those other modules, for not updating all the parts of config that it should do, when the module is installed).

### Checklist

- [ ] Tested in Development


BarryGeeMOJ (approved):
lgtm

PR #474, opened 2022-11-01
Disabling Sentry JS error reporting.

leonk:
### Context

No this is a small change.

### Intent

I'm trying to reduce noise on the #hub_alerts_prod channel.  We get quite a few clientside js errors from Drupal.  They are normally related to third party code (either from Drupal or ckeditor) and is not possible to replicate on another environment.  I.e. they are not possible to fix, and often don't actually represent a real issue.

I looked into a way to stop these types of errors from making it through to Slack (whilst still retaining PHP errors) but couldn't see a way.  I therefore think it's easier to just disable them altogether.  We've had them enabled for over a year and have never found them to be useful.

### Considerations

### Checklist

- [ ] Tested in Development

### DIFF
--- a/config/sync/raven.settings.yml
+++ b/config/sync/raven.settings.yml
@@ -18,7 +18,7 @@ message_limit: 2048
 trace: true
 fatal_error_handler: true
 fatal_error_handler_memory: 2560
-javascript_error_handler: true
+javascript_error_handler: false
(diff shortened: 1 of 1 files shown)


BarryGeeMOJ (approved):
lgtm

PR #475, opened 2022-11-01
Remove link summary field, which is not being used. 

leonk:
### Context

https://trello.com/c/7cqZ8r3j/1710-separate-summary-and-description-in-drupal

### Intent

Prep for the card above.

For the work on the card we need to create a new `field_summary`, but this is already in use on the link content type.

This PR removes the field, which we've found is not actually being used.  We can then re-create it on a subsequent PR.

### Considerations

### Checklist

- [ ] Tested in Development


BarryGeeMOJ (approved):
lgtm

PR #476, opened 2022-11-03
Adding make sync command to copy db down from s3 backups.

leonk:
### Context

https://trello.com/c/66Bk2qCz/1716-spike-improve-local-environment-setup

### Intent

Adds a `make sync` command for copying the latest db (from the s3 backups) and importing into the local db environment.

TODO: Add some documentation about this.

### Considerations

### Checklist

- [ ] Tested in Development

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -142,6 +142,9 @@ RUN composer install \
   --prefer-dist
 
 FROM test as local
+
+COPY scripts/ scripts/
+
 USER root
 RUN pecl install xdebug-3.1.5 \
   && docker-php-ext-enable xdebug
--- a/Makefile
+++ b/Makefile
@@ -29,3 +29,13 @@ deploy:
 	echo "Disabling maintenance and readonly mode"
 	drush state-set readonlymode_active 0
 	drush state-set system.maintenance_mode 0
+
+sync:
+	# Downloading latest db backup from S3
+	./scripts/downloadDBFromBackup.sh
+	# Download complete
+	# Clearing Drupal db of existing data
+	docker-compose -f ../prisoner-content-hub/docker-compose.yml exec -it prisoner-content-hub-backend drush sql-drop -y
+	# Importing db backup to Drupal DB
+	docker-compose -f ../prisoner-content-hub/docker-compose.yml exec prisoner-content-hub-backend scripts/importDBFromBackup.sh
+	# Import complete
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -52,10 +52,10 @@ jobs:
     <<: *defaults
     parameters:
       docker-compose-filename:
-        default: docker-compose.circleci.yml
+        default: docker-compose.yml
```
leonk, line 55: Circleci now uses the same docker-compose file as the local environment (so there's no need to manage to separate files).

.circleci/config.yml
```diff
@@ -52,10 +52,10 @@ jobs:
...
         type: string
       docker-image-name-backend:
-        default: prisoner-content-hub-backend
+        default: drupal
```
leonk, line 58: No need to prefix the image name anymore, as docker-compose will do that for us with the project name.  This means we get a docker name: `prisoner-content-hub-backend-drupal-1` instead of `prisoner-content-hub-backend-prisoner-content-hub-backend-1`

.gitignore
```diff
@@ -38,3 +37,8 @@ docroot/.eslintrc.json
 docroot/.ht.router.php
 docroot/.htaccess
 docroot/INSTALL.txt
+
+# Ignore override docker-compose files.
+# Even those these are in git, we want to allow local changes to them without being pushed up to git.
+docker-compose.override.yml
+prisoner-content-hub-backend-local.env
```
leonk, line 44: This might be a little bit weird to some. But basically I want to add in these files to git, with the default values that they come with.  But can see that developers may want to make additional changes to them, and not have them pushed back to git (or show up as a change in git).  So I've added these files in the PR, but also set them to be ignored.

phpunit.xml
```diff
@@ -13,7 +13,7 @@
 -->
 <phpunit bootstrap="vendor/weitzman/drupal-test-traits/src/bootstrap.php">
   <php>
-    <env name="DTT_BASE_URL" value="http://prisoner-content-hub-backend:8080"/>
+    <env name="DTT_BASE_URL" value="http://drupal:8080"/>
     <env name="DTT_API_URL" value="http://chrome:9222"/>
     <!-- <env name="DTT_MINK_DRIVER_ARGS" value='["chrome", { "chromeOptions" : { "w3c": false } }, "http://localhost:4444/wd/hub"]'/> -->
```
BarryGeeMOJ, line 18: Do we need the commented out code?

leonk: The xml file comes from Drupal (inc the commented code), so it’s there just to keep it as close to the original as possible.

also this PR isn’t modifying the commented part. If we wanted to removed it I think it should be in a separate PR

PR #486, opened 2022-12-14
Split summary field

leonk:
### Context

> Does this issue have a Trello card?

https://trello.com/c/7cqZ8r3j/1710-separate-summary-and-description-in-drupal
https://trello.com/c/KYWYItAd/1820-rename-description-to-main-body-content-on-basic-pages
https://trello.com/c/BygfBgWd/1889-add-character-limit-to-summary-field

### Intent

Changes in this PR:
- Hides `field_moj_description` (to be later removed after this has been deployed).
- Creates a separate `field_summary` on basic pages, video, audio, and pdf. Field is limited to 180 characters.
- Creates a `field_main_body_content` on basic page content type.
- Creates a `field_description` on audio and video content type.
- Field data migration in a deploy hook (to be deployed early in the morning, due to the cache clear that will occur).

### Considerations
This will need to stay in PR until https://trello.com/c/YmWkAKKy is done for the F/E.  And then both deployed at the same time.

> Is there any additional information that would help when reviewing this PR?

> Are there any steps required when merging/deploying this PR?

### Checklist

- [ ] This PR contains **only** changes related to the above card
- [ ] Tests have been added/updated to cover the change
- [ ] Documentation has been updated where appropriate
- [ ] Tested in Development

PR #499, opened 2023-01-17
Remove old description field

leonk:
### Context

Removing the old `field_moj_description` now that https://github.com/ministryofjustice/prisoner-content-hub-backend/pull/486 has been deployed.

### Intent

`field_moj_description` is no longer being used.  We kept it on the site so that both new and old fields would work during the last release.  Now that the f/e has switched to the new field name, we can remove this old one.

### Considerations

As this field is indexed in search, when it gets removed, drupal will mark the whole index as stale.  It will need to be rebuilt immediately (best to do this through the UI, so it can be done using batches).

### Checklist

- [ ] Tested in Development


composer.json
```diff
@@ -164,7 +164,8 @@
                 "Issue #3090131 - Use per-bundle entity list cache tags to potentially increase caching and performance": "patches/drupal-3090131-16.patch"
             },
             "drupal/views_ef_fieldset": {
-                "Issue #3173822: Exposed operators are not included in fieldsets": "patches/views_ef_fieldset_3173822-operator_issue-9.patch"
+                "Issue #3173822: Exposed operators are not included in fieldsets": "patches/views_ef_fieldset_3173822-operator_issue-9.patch",
+                "Issue #3269083: TypeError: Unsupported operand types:": "patches/views_ef_fieldset-3269083-17.patch"
```
leonk, line 168: Had to add in this patch as the views_ef_fieldset module was broken, and I couldn't arrange the new description field filter on the /admin/content page

psoleckimoj (approved):
LGTM!

BarryGeeMOJ (approved):
LGTM

PR #507, opened 2023-03-23
Feature/linkit suggestions for front end urls

Eli-TW:
### Context

https://trello.com/c/8M9EFLQK/2083-add-non-drupal-suggestions-to-linkit

- Start creating a basic page. 
- Add a link into the main body content
- Start typing a link for a front end URL that doesn't exist within Drupal, for example /profile

Before this change the UI would not acknowledge this as a valid link. This could confuse users.

### Intent

A new custom module is added, that contains a custom Linkit matcher plugin to match against the following paths
- /profile
- /timetable
- /money/transactions
- /approved-visitors

A partial case insensitive match against any of these paths or their descriptions will trigger a linkit suggestion in a new group 'Profile links'. This group is positioned above the suggestions from Drupal's own content, to make sure it doesn't get buried.

The new module is enabled in config, and the new plugin added to the default linkit configuration.

This change also updates the version of the linkit module we are using. This isn't strictly necessary for this change, but we do need to update this module for Drupal 10 compatibility, and this is a convenient time to do so.

Yes! Here are some.

Before this change

image

After the change

image

### Considerations

No

No

### Checklist

- [x] Tested in Development

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3582,29 +3582,30 @@
         },
         {
             "name": "drupal/linkit",
-            "version": "6.0.0-beta3",
+            "version": "6.0.0-beta4",
             "source": {
                 "type": "git",
                 "url": "https://git.drupalcode.org/project/linkit.git",
-                "reference": "6.0.0-beta3"
+                "reference": "6.0.0-beta4"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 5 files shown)


psoleckimoj (approved):
LGTM!

PR #514, opened 2023-04-12
⬆️   upgrade from CKEditor 4 to 5. Reconfigure all text formats to use CKEditor 5.

Eli-TW:
### Context

https://trello.com/c/3VN3zEKj/2-ckeditor

N/A

### Intent

* Uninstalls CKEditor Drupal module (which wraps CKEditor 4) and installs CKEditor 5 module.
* Reconfigures Basic HTML and Full HTML text formats to use CKEditor 5.

Sure!

Old CKEditor 4 basic HTML editor

image

Shiny new CKEditor 5 basic HTML editor

image

Example CKEditor 4 front end output

image

Identical CKEditor 5 front end output

image

Full HTML is similarly identical in functionality and output between CKEditor 4 and 5.

### Considerations

No.

We need to inform the DCMs that the editing experience will look different, and in some cases will have slightly different flows, for example when inserting images / tables etc.

The new flows should be intuitive enough that I don't think producing documentation around it is necessarry. However we should be on hand to support in Teams should anyone need assistants.

### Checklist

- [ ] Tested in Development

### DIFF
--- a/config/sync/editor.editor.basic_html.yml
+++ b/config/sync/editor.editor.basic_html.yml
@@ -1,49 +1,63 @@
-uuid: 660d223e-2fbf-4f9a-830e-7411b8f4f663
+uuid: 7ccf4663-8641-4c29-9cb1-4c5fc07e1b3f
 langcode: en
 status: true
 dependencies:
   config:
     - filter.format.basic_html
   module:
-    - ckeditor
-_core:
-  default_config_hash: AqlPmO16LvJI4D0Ih6u4GFQIzqr5OnLgAUSjcUGWk2g
+    - ckeditor5
 format: basic_html
-editor: ckeditor
+editor: ckeditor5
 settings:
   toolbar:
-    rows:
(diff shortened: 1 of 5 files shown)


config/sync/editor.editor.basic_html.yml
```diff
@@ -1,49 +1,63 @@
...
+      - numberedList
+      - '|'
+      - blockQuote
+      - drupalInsertImage
+      - '|'
+      - heading
+      - '|'
+      - sourceEditing
+      - '|'
+      - code
```
psoleckimoj, line 29: Do we want to add code in as it was never previously there and inexperienced DCMs could create a lovely mess of HTML and end up asking us what's wrong with the page?

psoleckimoj: Ignore me!!

psoleckimoj (approved):
LGTM!

PR #528, opened 2023-05-16
🔧  switch temp sync region to be Ireland, so we can sync new dev bucket from old prod. 

Eli-TW:
This will keep the new dev bucket in sync with changes made by content editors until we switch to the new bucket.

### Context

https://trello.com/c/IaJKTzZH/2128-move-prod-s3-bucket-from-ireland-to-london

N/A

### Intent

There is a temporary sync cron job that was used to sync the new dev bucket from the existing staging bucket. Now this sync is complete, we want to stay up to date with any further changes made by content editors, which are still using the original Irish S3 prod bucket. This PR changes the region from which we run the temporary sync to Ireland.

No.

### Considerations

No

Yes - the k8s secret drupal-s3-output-temp needs to be updated to point to the original prod s3 bucket at the same time. It currently points at the original staging bucket.

### Checklist

- [ ] Tested in Development

### DIFF
--- a/helm_deploy/prisoner-content-hub-backend/values.development.yaml
+++ b/helm_deploy/prisoner-content-hub-backend/values.development.yaml
@@ -18,6 +18,7 @@ s3Sync:
 # Temporary switch for syncing new bucket from staging.
 s3SyncTemp:
   enabled: true
+  source_region: eu-west-1
 
 application:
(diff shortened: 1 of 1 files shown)


psoleckimoj (approved):
lgtm

PR #529, opened 2023-05-17
Update s3 temp sync region in prod

Eli-TW:
### Context

https://trello.com/c/IaJKTzZH/2128-move-prod-s3-bucket-from-ireland-to-london

N/A

### Intent

There is a temporary sync job. This previously sync'd the new prod s3 bucket in London with the staging bucket, to get the majority of the contents from the same region. Now that has been completed, we now want to sync the new prod bucket from the old one which is still use by the prod application. Then when we want to cut over to the new prod bucket it will already be up to date.

No

### Considerations

No

Yes - the drupal-s3-output-temp secret in the prod namespace will need updating to reflect the old prod bucket.

### Checklist

- [x] Tested in Development


psoleckimoj (approved):
LGTM!

PR #582, opened 2024-03-18
LNP-580: core & contrib updates

Eli-TW:
### Context

https://dsdmoj.atlassian.net/browse/LNP-580

N/A

### Intent

Updates core and contrib modules. Removes/rerolls patches as necessary.

No

### Considerations

No

A follow is required to remove the config_filter module. This was previously an implicit requirement for this project via the dependencies of the config_ignore module. This PR updates the major version of config_ignore that no longer requires config_filter, but we cannot remove config_filter from the filesystem whilst it is still enabled in Drupal.

Therefore this PR adds config_filter as an explicit composer dependency so it is not removed from the filesystem in this PR, but does uninstall it by removing it from core.extension.yml.

After this deployment has run on production, we need a follow up PR that removes the explicit composer dependency on config_filter mentioned above.

### Checklist

- [x] Tested in Development

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "drupal/field_group": "^3.1",
         "drupal/file_mime_validator": "^2.0",
         "drupal/flood_control": "^2.3",
-        "drupal/flysystem": "dev-2.2.x",
+        "drupal/flysystem": "^2.2",
         "drupal/flysystem_s3": "^2",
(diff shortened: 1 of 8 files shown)


psoleckimoj (approved):
LGTM!

sweetymj (approved):
Looks good.

PR #605, opened 2024-07-18
Story/LNP-820 recently added hq priority content

Eli-TW:
### Context

https://dsdmoj.atlassian.net/browse/LNP-820

N/A

### Intent

* Introduces a new role exclusively for the comms team
* Creates a new boolean field to prioritise the 5 main prisoner facing content types
* Restricts access to that field to admin and those with the new comms role
* Defaults the priority field to be set to true when members of the comms role create content
* Adjusts the Recently Added JSON:API resource, and hence the Recently Added section of homepages, to return half prioritised content, half unprioritised content, with the prioritised content showing first. 

No

### Considerations

No

No

### Checklist

- [x] Tested in Development

This PR does include some changes not directly associated with the story; since upgrading to PHP 8.3, use of undeclared class properties became deprecated. This raises a lot of noise in test runs, and as we were writing new tests, it was appropriate to add property declarations to remove the deprecation warnings.

Also there was duplicated code in the test classes for a function getJsonApiResponse(). As we needed something similar for the new tests, this was refactored into a trait and the duplicate code removed from all test classes, rather than duplicate the code one more time and exacerbating the problem.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -192,6 +192,9 @@
             },
             "drupal/search_api": {
                 "Issue #3313569: Warning: Could not load the following items on index": "patches/3313569-10--referenced_entity_tracking_translations.patch"
+            },
+            "drupal/field_permissions": {
+                "Issue #3411976: 500 error with JSONAPI on Base Entity Fields": "patches/file_permissions.3411976-25.patch"
             }
         },
         "patches-ignore": {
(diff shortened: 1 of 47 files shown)


sweetymj (dismissed):
LGTM.

psoleckimoj (approved):
LGTM!

psoleckimoj (commented):
LGTM, but tbh too much for me to review. I can usually work stuff out when a PR is related to one thing even if I don't know the language so well. Probably better for @sweetymj to approve properly.

sweetymj (approved):
LGTM.

PR #611, opened 2024-08-14
LNP-718: 🐛  fix bug in recently-added endpoint

Eli-TW:
trim priority content to half the required value before loading non-priority content. This fixes the issue where non-priority content tries to load 0 entries and then crashes.

Also return early if trying to load 0 items in the recently-added endpoint. This can still happen despite the above if the requested page size is 1 - this leads to 1 item of priority content (if there is any) and 0 items of non-priority content.

### Context

https://dsdmoj.atlassian.net/browse/LNP-952

Prioritise lots of content. Attempt to load a prison home page.

### Intent

Trims the size of the set of prioritised content before working how much non-prioritised content to load.
This protects against against trying to load 0 non-prioritised items which was the condition that triggered the error.

However it is also possible to trigger the error when the end point is called with a size limit of 1 if there is 1 piece of prioritised content. Therefore both load functions in the class return early when asked to load 0 items.

No

### Considerations

No

No

### Checklist

- [x] Tested in Development

### DIFF
--- a/docroot/modules/custom/prisoner_hub_recently_added/src/Resource/RecentlyAdded.php
+++ b/docroot/modules/custom/prisoner_hub_recently_added/src/Resource/RecentlyAdded.php
@@ -112,6 +112,12 @@ class RecentlyAdded extends EntityResourceBase implements ContainerInjectionInte
       return $b['published_at'] <=> $a['published_at'];
     });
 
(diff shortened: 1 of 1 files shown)


psoleckimoj (approved):
LGTM!

PR #612, opened 2024-08-16
LNP-718: content in multiple places

Eli-TW:
### Context

https://dsdmoj.atlassian.net/browse/LNP-718

N/A

### Intent

* Removes the Field Not In Series checkbox - this is redundant as you can tell whether a field is in a series depending on if you select a series. This removed field was used to show/hide the Category field, because prior to this PR, content could only be in either a series or a category
* Removes the dependency on the contrib module field_state_empty, and uninstalls it from the site. This was only used to clear out other fields in response to the state of the Field Not In Series checkbox, which is now removed (see above). Note this doesn't remove this module from the composer dependencies, that cannot be done in the same deploy during which it will be uninstalled, and this will be done in a follow up.
* Updates the form help text for the 5 main content types to reflect content can now have a series and up to 3 categories.
* Updates the configuration of the category field to allow three categories.
* Adds a constraint to make sure that content of the 5 main content types has a series or a category
* Updates all tests that create nodes hitting this constraint to automatically add a category, so that the tests don't fail due to the new constraint
* Removes tests that would hit the constraint and it's inappropriate to fix
* Add a presave function to remove series data (eg season/release date/episode number) when saving content with no series, rather than using the field_not_in_series field and form_state_empty.

No

### Considerations

No

No

### Checklist

- [ ] Tested in Development


psoleckimoj (approved):
Sorry, forgot to approve. LGTM!

PR #613, opened 2024-08-21
LNP-968: 🔥  drop target database before importing database dump.

Eli-TW:
### Context

https://dsdmoj.atlassian.net/browse/LNP-968

### Intent

Adds flags to the mysqldump command run to take exports from production, so that when the dump is imported into other environments, the existing database will be dropped and recreated before the exported data is imported.

This ensures that any tables that exist eg in dev but not prod are removed on import. This stops issues where for example Drupal tries to create a table in a post-install action and fails because it unexpectedly exists.

No

### Considerations

No

No

### Checklist

- [ ] Tested in Development

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,8 +41,6 @@ sync:
 	# Downloading latest db backup from S3
 	docker-compose exec drupal scripts/downloadDBFromBackup.sh
 	# Download complete
-	# Clearing Drupal db of existing data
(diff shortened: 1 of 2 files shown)


psoleckimoj (approved):
LGTM!

psoleckimoj (approved):
LGTM!

PR #614, opened 2024-09-04
LNP-975 account for different db names when refreshing db

Eli-TW:
### Context

https://dsdmoj.atlassian.net/browse/LNP-975

Try to run make-sync locally. The script will try to delete a database with the same name as the prod database, and fail.

### Intent

In LNP-968, we addressed the problem that during a database refresh, tables that were on the target system but not in the database dump would still be present after refresh.

That change built the dropping and recreation of the target database into the dump file. However, because that embeds the database name, and the database name is different across environments, this would not work.

This PR refactors that to move the dropping and recreation of the target database to the db-refresh.sh script for AWS environments, and to the Makefile for local environments.

No

### Considerations

No

No

### Checklist

- [ ] Tested in Development

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,8 @@ sync:
 	# Downloading latest db backup from S3
 	docker-compose exec drupal scripts/downloadDBFromBackup.sh
 	# Download complete
+	# Clearing Drupal db of existing data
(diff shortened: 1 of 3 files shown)


madebyzak (approved):
LGTM

PR #623, opened 2024-11-06
LNP-1047: 🛂  remove permissions from administrator role that were mov…

Eli-TW:
…ed into the Actions UI module (that we do not have enabled). See https://www.drupal.org/node/3413949. This can cause a fatal error with some permission operations, for example, when uninstalling the toolbar module.

### Context

https://dsdmoj.atlassian.net/browse/LNP-1047

Uninstall the toolbar module. Fatal error happens.

### Intent

Removes two permissions from the administrator role. These permissions have been moved to a module that we do not have enabled, and therefore when recalculating permissions can cause the site to crash.

No

### Considerations

No

No

### Checklist

- [x] Tested in Development


psoleckimoj (approved):
LGTM!

